PR #11, opened 2017-09-16
Changed documentation to better explain installation process

rageycomma:
I added a bit more detail so that those with no experience of typings can install this module and get started instantly.


bryanmacfarlane:
@rageycomma - thanks for your contribution.  however, we have another PR which is getting rid of the deprecated typings. 

https://github.com/Microsoft/typed-rest-client/pull/12

PR #12, opened 2017-10-23
Users/stfrance/handle redirects

stephenmichaelf:
This PR includes:
- Handling redirects (for GET only, I used the same defaults as .NET framework, it's enabled with 50 redirects unless you override)
- Replacing Typings with @ types in both the lib and samples
- Fixing environment variable usage(had issues with the existing code, may be because I increased node version. The type for env vars is a dictionary now.)


bryanmacfarlane:
@stephenmichaelf @ericsciple - should the redirect loop be in the request lower function so all verbs just get it?  Also, shouldn't I be able to specify this as options in the http client constructor (it takes an option structure).  I want to avoid dirtying up all the individual methods with tons of args going forward.  I would expect options to default correctly if not specfied.

Finally, what about the rest client (which calls the http client).  Shouldn't it pass defaults through (or take in it's options and pass through)?

Let's get it "right" and not revisit.

I want to put some focus back on this and get it done today / monday.

lib/HttpClient.ts
```diff
@@ -124,8 +129,22 @@ export class HttpClient {
         return this.request('OPTIONS', requestUrl, null, additionalHeaders || {});
     }
 
-    public get(requestUrl: string, additionalHeaders?: ifm.IHeaders): Promise<HttpClientResponse> {
-        return this.request('GET', requestUrl, null, additionalHeaders || {});
+    public async get(requestUrl: string, additionalHeaders?: ifm.IHeaders, allowRedirects: boolean = true, maxRedirects: number = 50): Promise<HttpClientResponse> {
+        let response = await this.request('GET', requestUrl, null, additionalHeaders || {});
+
+        while (response.message.statusCode === HttpCodes.ResourceMoved
```
bryanmacfarlane: Shouldn't we handle all of these? 

    MovedPermanently = 301,
    ResourceMoved = 302,
    TemporaryRedirect = 307,
    PermanentRedirect = 308,

stephenmichaelf: I wasn't sure about this but it makes sense to handle them all. I added this change.

stephenmichaelf:
(discussed offline, notes for posterity)

1. should the redirect loop be in the request lower function so all verbs just get it? 

This seems risky to me given that the other verbs make data changes but I checked .NET Core and from what I can see they do allow redirects for all verbs so we probably should too.

2. Also, shouldn't I be able to specify this as options in the http client constructor (it takes an option structure).

Agreed. Will move.

3. Finally, what about the rest client (which calls the http client). Shouldn't it pass defaults through (or take in it's options and pass through)?

Agreed. Will pass through and use defaults from HttpClient if they aren't passed in rest client.

stephenmichaelf:
@bryanmacfarlane Eric and I reviewed this offline, it should be good to go.

FYI:
- This includes redirects for all HTTP verbs as well as replaces Typings with @ types
- The minor version has been bumped from 12 to 13. Technically this is a change in functionality(redirects are allowed as default now) but Eric and I agreed it seems like overkill to bump the major, especially since we are at 0 for the major version at the moment

PR #13, opened 2017-10-26
Added simple export for handlers

tasadar2:
Usage: `import * as httphandlers from "typed-rest-client/Handlers"`


bryanmacfarlane:
@stephenmichaelf - can you review?

stephenmichaelf:
@tasadar2 Thank you for your contribution! We are having some discussions about the best way to approach this and will have a write up here as we get a better understanding of how consumers can use the http handlers.

PR #18, opened 2017-11-19
Add process.exit to samples.

stephenmichaelf:
This resolves the issue of the sample run hanging.

### DIFF
--- a/samples/samples.ts
+++ b/samples/samples.ts
@@ -4,6 +4,7 @@ import * as restSamples from './rest';
 async function run() {
     await httpSamples.run();
     await restSamples.run();
+    process.exit();
 }
 
 run();


ericsciple:
have you tried divide and conquer to determine if a particular sample is hanging the process? I wonder if a promise is never finishing

bryanmacfarlane:
This seems like a workaround.  Do we understand why it's hanging (and never did before).  What's more concerning is our tasks are written with this pattern so if this is new in node 8, we need to figure it out.  process.exit is a no no in tasks because there were cases where it would kill the process before stderr would drain with that last set result fail missing was pretty important :)

bryanmacfarlane:
I also like @ericsciple suggestion of divide and conquer.  I was thinking of trying before the last set of commits.  And then trying combinations with node 6 and node 8 (or previous versions of node 8).

stephenmichaelf:
I started doing divide and conquer, at first it seemed like it was hanging on 'get request and pipe stream' but now it seems to be 'get request with implicitly allowed redirects'. It could be that I made a subtle mistake during my testing that led me in the wrong direction.

The strange thing is that 'get request with implicitly allowed redirects' is still returning the redirected http response code and body. Which also means the initial 302 is returning fine. I wonder if something is going on inside _requestRaw.

Continuing to dig.

I just ran it with node 6.12.0 and it returned immediately... on master. I am going to move up the versions until it hangs.

stephenmichaelf:
Scratch that, I was on my branch with process.exit :)

stephenmichaelf:
Running from hash 8d763f7b75a57fdb8c9f38ad4bad9f50acaf2514

Node 6.12.0 -> fails to build
Node 8.1.0 -> fails to build
Node 8.9.1 -> fails to build

Failing to find namespaces for NodsJS, I think it's an issue with Typings.

The hang only seems to be happening when we go inside the while loop for redirects. I wonder if it's a promise inside _requestRaw hanging or if the scoping of variables gets messed up.

I also used [wtfnode](https://www.npmjs.com/package/wtfnode) and got this:

[WTF Node?] open handles:
- File descriptors: (note: stdio always exists)
  - fd 1 (tty) (stdio)
  - fd 2 (tty) (stdio)
- Sockets:
  - 192.168.1.68:62403 -> 54.225.152.117:80

Going to take a break, I am probably missing something simple here.

bryanmacfarlane:
Yeah, 54.225.152.117:80 is httpbin.org so a socket is left open and node won't exit.

bryanmacfarlane:
I have a solution but still not 100% positive it's the "right fix".  But it works.  It does however prove my point above that it's a lingering socket.  Theory is because of keep alives

https://github.com/Microsoft/typed-rest-client/pull/19

PR #19, opened 2017-11-19
redir hang

bryanmacfarlane:


stephenmichaelf:
Strange to me that it doesn't hang with 1 request but does on the redirect...

bryanmacfarlane:
I think I understand it now.  Will go over it with you tomorrow

Get Outlook for iOS<https://aka.ms/o0ukef>

stephenmichaelf:
I think I see what you're saying, we never call ReadBody on the first response... Sounds good.

bryanmacfarlane:
Yeah, and we shouldn't.  I think we need to readbody on the one we throw away ( you reassign the car ) - or just call destroy but that has the downside of dropping the socket from the server instead of cleanly finishing the request.  Note the first call returns the stream so the call calls read body.  Of course they don't on the one we throw away on retries

I will confirm when I'm home soon

Get Outlook for iOS<https://aka.ms/o0ukef>

bryanmacfarlane:
@ericsciple and @stephenmichaelf - updated with a cleaner fix

lib/HttpClient.ts
```diff
@@ -177,20 +177,26 @@ export class HttpClient {
...
-          }
+            const redirectUrl: any = response.message.headers["location"];
+            if (!redirectUrl) {
+                // if there's no location to redirect to, we won't
+                break;
+            }
+
+            // we need to finish reading the response before reassigning response
+            // which will leak the open socket.
+            response.readBody();
```
ericsciple: is this synchronous or are you missing an await here? otherwise looks fine to me

bryanmacfarlane: Ahhh - great catch 🥇  .  It worked because it probably only needed to read one block off the stream and it wasn't the last test in the sample

PR #23, opened 2017-11-21
Support path in the base url

kaeedo:
This PR adds the option to include an option to preserve the pathname specified in the base URL for the Rest Client. Default is false (maintains old behavior)

### DIFF
--- a/lib/Util.ts
+++ b/lib/Util.ts
@@ -5,10 +5,9 @@ import * as path from 'path';
  * creates an url from a request url and optional base url (http://server:8080)
  * @param {string} requestUrl - a fully qualified url or relative url
  * @param {string} baseUrl - an optional baseUrl (http://server:8080)
- * @param {boolean} preservePath - an optional baseUrl (http://server:8080)
  * @return {string} - resultant url 
  */
-export function getUrl(requestUrl: string, baseUrl?: string, preservePath?: boolean): string  {
+export function getUrl(requestUrl: string, baseUrl?: string): string  {
     if (!baseUrl) {
         return requestUrl;
     }
@@ -21,9 +20,7 @@ export function getUrl(requestUrl: string, baseUrl?: string, preservePath?: bool
     combined.auth = combined.auth || base.auth;
     combined.host = combined.host || base.host;
 
-    if (preservePath) {
-        combined.pathname = path.join(base.pathname, combined.pathname);
-    }
+    combined.pathname = path.join(base.pathname, combined.pathname);
 
     // path from requestUrl always wins
     let res: string = url.format(combined);
--- a/lib/Util.ts
+++ b/lib/Util.ts
@@ -28,7 +28,5 @@ export function getUrl(requestUrl: string, baseUrl?: string, preservePath?: bool
     // path from requestUrl always wins
     let res: string = url.format(combined);
 
-    console.log(res)
-
     return res;
 }
--- a/lib/Util.ts
+++ b/lib/Util.ts
@@ -1,4 +1,5 @@
 import * as url from 'url';
+import * as path from 'path';
 
 /**
  * creates an url from a request url and optional base url (http://server:8080)
@@ -21,11 +22,13 @@ export function getUrl(requestUrl: string, baseUrl?: string, preservePath?: bool
     combined.host = combined.host || base.host;
 
     if (preservePath) {
-        combined.pathname = base.pathname + combined.pathname;
+        combined.pathname = path.join(base.pathname, combined.pathname);
     }
 
     // path from requestUrl always wins
     let res: string = url.format(combined);
+
+    console.log(res)
 
     return res;
 }
--- a/lib/Util.ts
+++ b/lib/Util.ts
@@ -20,8 +20,8 @@ export function getUrl(resource: string, baseUrl?: string): string  {
     resultantUrl.auth = resultantUrl.auth || base.auth;
     resultantUrl.host = resultantUrl.host || base.host;
 
-    let basePathComponent: string = base.pathname === '/' ? '': base.pathname;
-    resultantUrl.pathname = path.posix.join(basePathComponent, resultantUrl.pathname);
+    let basePathComponent: string = base.pathname === '/' ? '' : base.pathname;
+    resultantUrl.pathname = path.posix.resolve(basePathComponent, resultantUrl.pathname);
 
     let res: string = url.format(resultantUrl);
 
--- a/lib/Util.ts
+++ b/lib/Util.ts
@@ -20,9 +20,10 @@ export function getUrl(resource: string, baseUrl?: string): string  {
     resultantUrl.auth = resultantUrl.auth || base.auth;
     resultantUrl.host = resultantUrl.host || base.host;
 
-    let basePathComponent: string = base.pathname === '/'? '': base.pathname;
-    resultantUrl.pathname = path.resolve(basePathComponent, resultantUrl.pathname);
+    let basePathComponent: string = base.pathname === '/' ? '': base.pathname;
+    resultantUrl.pathname = path.posix.join(basePathComponent, resultantUrl.pathname);
 
     let res: string = url.format(resultantUrl);
+
     return res;
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,13 +1,17 @@
...
+  "version": "0.15.0",
   "description": "Node Rest and Http Clients for use with TypeScript",
   "main": "./RestClient.js",
   "scripts": {
     "build": "node make.js build",
     "test": "node make.js test",
     "samples": "node make.js samples"
   },
+  "engines": {
+    "node": ">=8.9.0",
```
bryanmacfarlane: node 8.9 / npm 5.1 is only a developer build time dependency.  At runtime this should and needs to work on 6.x LTS and even back to node 5.10.  I would remove this from the PR and make another issue to define runtime req (I want our CI to multi plex and validate on ideally >= node 4.x LTS but at least node >=6.x LTS)

lib/Util.ts
```diff
@@ -18,8 +19,13 @@ export function getUrl(requestUrl: string, baseUrl?: string): string  {
     combined.protocol = combined.protocol || base.protocol;
     combined.auth = combined.auth || base.auth;
     combined.host = combined.host || base.host;
-    // path from requestUrl always wins
 
+    if (preservePath) {
+        combined.pathname = base.pathname + combined.pathname;
```
bryanmacfarlane: path.join?  I think there is also interesting test cases with base not / having trailing slash and pathname not/having leading slash.

lib/Util.ts
```diff
@@ -4,9 +4,10 @@ import * as url from 'url';
  * creates an url from a request url and optional base url (http://server:8080)
  * @param {string} requestUrl - a fully qualified url or relative url
  * @param {string} baseUrl - an optional baseUrl (http://server:8080)
+ * @param {boolean} preservePath - an optional baseUrl (http://server:8080)
  * @return {string} - resultant url 
  */
-export function getUrl(requestUrl: string, baseUrl?: string): string  {
+export function getUrl(requestUrl: string, baseUrl?: string, preservePath?: boolean): string  {
```
bryanmacfarlane: I'm struggling with whether this should be the default.  In the normal case of just providing the hostname portion (no path) it does the right thing.  In the case where the base url has a relative path, it does the right thing.  I need to think about it.  

This is pre-release about to go to 1.0 and this change should bump the minor version (so ^ with patch level won't drift into this behavior) and we can doc.

The other thought was flip to opt in to chop the relative path but if that's your intent, just provide the hostname portion.  No need for another optional arg ...hmmmm

Thoughts?

test/resttests.ts
```diff
@@ -84,7 +85,93 @@ describe('Rest Tests', function () {
             assert(err['statusCode'] == 500, "statusCode should be 500");
             assert(err.message && err.message.length > 0, "should have error message");
         }
-    });    
+    });
+
```
bryanmacfarlane: Add divider

//--------------------------------------------------------
// Path tests
//--------------------------------------------------------

lib/Util.ts
kaeedo: Since this is still prerelease, my preference would be to never strip the relative path. Reason being similar to your thought about flipping to opt in. If the use case comes up that the base URL also has a path, then it would make sense to explicitly state it in the baseUrl parameter, and always expect it to be there. At that point the optional parameter can be completely removed. That would make this a breaking change though...

lib/Util.ts
kaeedo: Using `path.join` now. Also fixes the trailing and leading slash problems

test/resttests.ts
kaeedo: Done

lib/Util.ts
stephenmichaelf: @kaeedo Could you add some examples of usage for this feature? Then we can speak to the explicit use cases. It often helps think things through and make the "right" choice more apparent. (based on the perspective of a consumer of the library). As well as what the consumer would have to do if we didn't have this feature.

It's tricky to balance where we draw the line of what's included and what isn't.

bryanmacfarlane: @kaeedo - I agree on your points.  Bump the minor version with this change.  That mitigates the pre-release breaking change since folks default to sliding on patch versions.  And it's only a change when you're doing something non mainstream (passing base url with path but then expecting it not to be honored).  Let's move forward without the arg.  Add the test cases on trailing and leading slashes.

kaeedo: @stephenmichaelf 
For the use case in my office, we have a reverse proxy configured to point to different docker containers:

    https://www.contoso.com/firstcontainer
    https://www.contoso.com/secondcontainer
    https://www.contoso.com/thirdcontainer

I would like to create an instance of a `RestClient` such that `https://www.contoso.com/firstcontainer` is the `baseUrl` instead of just `https://www.contoso.com` and then having to prepend `firstcontainer` to every rest request

kaeedo:
Preserving path is now the default, and cannot be turned off. the version in `package.json` has been bumped to 0.15.0. Unit test for trailing slashes and no slashes are also there

bryanmacfarlane:
please resolve conflicts (bump version to 0.16) and I'll review tomorrow.

lib/Util.ts
```diff
@@ -18,8 +19,11 @@ export function getUrl(requestUrl: string, baseUrl?: string): string  {
     combined.protocol = combined.protocol || base.protocol;
     combined.auth = combined.auth || base.auth;
     combined.host = combined.host || base.host;
-    // path from requestUrl always wins
 
+    combined.pathname = path.join(base.pathname, combined.pathname);
```
bryanmacfarlane: Instead of path.join, let's try path.resolve.
So 
'http://httpbin.org' and 'foo/bar' = 'http://httpbin.org/foo/bar'
'http://httpbin.org' and '/foo/bar' = 'http://httpbin.org/foo/bar'
'http://httpbin.org/baz' and 'foo/bar' = 'http://httpbin.org/baz/foo/bar'
'http://httpbin.org/baz' and '/foo/bar' = 'http://httpbin.org/foo/bar'

I sketched it out in a sketch PR. 

https://github.com/Microsoft/typed-rest-client/pull/28

See my tests.  You can merge in with your PR.  

Also note I bumped to 1.0 and will release with tag of preview.  After thinking about it, I think that's the best thing to do.

Thoughts?

kaeedo: The problem with using `path.resolve` on a Windows system, is that an absolute path will include a drive letter. For the test case

`let res: string = util.getUrl('get/foo', 'http://httpbin.org/bar');`
`res` comes out to `http://httpbin.org/C:\bar\get\foo`

bryanmacfarlane:
not a problem - if you do something illogical like putting a drive letter in an url path it won't work.  We're optimizing for the positive case scenarios (rooted and unrooted relative paths)


Sent from Outlook<http://aka.ms/weboutlook>

test/resttests.ts
```diff
@@ -176,5 +177,27 @@ describe('Rest Tests', function () {
         assert(restRes.result.args.baz === 'top');
     });
 
+    //
+    // getUrl path tests
```
bryanmacfarlane: you're missing this test.

'/get/foo', 'http://httpbin.org/bar'

should result in 'http://httpbin.org/get/foo'

lib/Util.ts
```diff
@@ -3,27 +3,27 @@ import * as path from 'path';
...
-    combined.host = combined.host || base.host;
+    // resource (specific per request) eliments take priority
+    let resultantUrl: url.Url = url.parse(resource);
+    resultantUrl.protocol = resultantUrl.protocol || base.protocol;
+    resultantUrl.auth = resultantUrl.auth || base.auth;
+    resultantUrl.host = resultantUrl.host || base.host;
 
-    combined.pathname = path.join(base.pathname, combined.pathname);
+    let basePathComponent: string = base.pathname === '/' ? '': base.pathname;
+    resultantUrl.pathname = path.posix.join(basePathComponent, resultantUrl.pathname);
```
bryanmacfarlane: use resolve.  drive letter on windows not a problem.  If you do something illogical like using a drive letter on windows, you will get a failure which is expected.  We're optimizing for the positive rooted and unrooted cases.

kaeedo: I see what it's doing now.

bryanmacfarlane:
Thanks!

bryanmacfarlane:
published as typed-rest-client@preview

Centro1993:
Why is this not in Master? Just discovered that BaseUrl cannot be a relative path. Worked fine for development, but now that I want to take my project online, I want to redirect all my API Calls to /api.

PR #24, opened 2017-11-28
Downgrade node version dev dependency.

stephenmichaelf:
Downgrade node version dev dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typed-rest-client",
-  "version": "0.14.2",
+  "version": "0.15.0",
   "description": "Node Rest and Http Clients for use with TypeScript",
   "main": "./RestClient.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -27,10 +27,10 @@
   "homepage": "https://github.com/Microsoft/typed-rest-client#readme",
   "devDependencies": {
     "@types/mocha": "^2.2.44",
-    "@types/node": "^8.0.19",
+    "@types/node": "^6.12.0",
     "@types/shelljs": "0.7.4",
     "mocha": "^3.5.3",
-    "node": "^8.0.19",
```
bryanmacfarlane: I understand the @types/node above, but what purpose does this line address?

stephenmichaelf: I thought we could use it to require a certain version of node for development? I know there is an "engines" option but I believe that is for execution and not development (though we may want to add that too). Is it sufficient to only reference node using the @ type?

bryanmacfarlane: Yeah, I don't think it works that way.  If you want to do that, you should do it in the make.js file.  Let's remove that line and then do a separate PR for that.

Actually, I'm going to merge this and then do a checkin that removes it.

PR #30, opened 2017-12-04
Support http keep alives

omeshp:


omeshp:
29

lib/HttpClient.ts
```diff
@@ -297,48 +297,53 @@ export class HttpClient {
         info.options.method = method;
         info.options.headers = headers || {};
         info.options.headers["User-Agent"] = this.userAgent;
+        // if an agent is supplied in options use that agent for managing connections
+        if (!!this.requestOptions && !!this.requestOptions.agent) {
+            info.options.agent = this.requestOptions.agent;
```
bryanmacfarlane: I think it's odd that if you supply an agent, the other options like proxy config and maxSockets are ignored.  That leaves a couple of options (1) layer those onto the agent supplied in specified in options (which means refactoring this flow in this function) or (2) simply add another option (or make it default) to re-use the agent out code creates.

The second option might be the best since it offers the performance benefit of re-using the socket without offloading the responsibilities of setting up the proxy etc... to the consumer.

Thoughts?

lib/HttpClient.ts
```diff
@@ -257,8 +260,52 @@ export class HttpClient {
...
+    private _getRequestAgent(requestUrl) {
+        let agent;
+        let keepAlive = false;
+        let maxSockets = http.globalAgent.maxSockets;
+
+        if (!!this.requestOptions) {
+            keepAlive = !!this.requestOptions.keepAlive;
+            maxSockets = this.requestOptions.maxSockets || http.globalAgent.maxSockets
+        }
+
```
GitHubSriramB: remove blank line

lib/HttpClient.ts
```diff
@@ -257,8 +260,52 @@ export class HttpClient {
...
+        return info;
+    }
+
+    private _getRequestAgent(requestUrl) {
+        let agent;
+        let keepAlive = false;
+        let maxSockets = http.globalAgent.maxSockets;
+
+        if (!!this.requestOptions) {
+            keepAlive = !!this.requestOptions.keepAlive;
```
GitHubSriramB: add destroy agent support to make sure connections get closed after use

lib/HttpClient.ts
```diff
@@ -257,8 +260,52 @@ export class HttpClient {
...
+        info.httpModule = usingSsl ? https : http;
+        info.options = <http.RequestOptions>{};
+        info.options.host = info.parsedUrl.hostname;
+        info.options.port = info.parsedUrl.port ? parseInt(info.parsedUrl.port) : defaultPort;
+        info.options.path = (info.parsedUrl.pathname || '') + (info.parsedUrl.search || '');
+        info.options.method = method;
+        info.options.headers = headers || {};
+        info.options.headers["User-Agent"] = this.userAgent;
+        
+        let agent;
```
GitHubSriramB: keep all this logic within _getRequestAgent

lib/HttpClient.ts
```diff
@@ -257,8 +260,52 @@ export class HttpClient {
...
+        if (this.handlers) {
+            this.handlers.forEach((handler) => {
+                handler.prepareRequest(info.options);
+            });
+        }
+
+        return info;
+    }
+
+    private _getRequestAgent(requestUrl) {
```
GitHubSriramB: _getAgent is better?

lib/HttpClient.ts
omeshp: 2nd suggestion looks good to me. I have refactored the method to reuse agent across requests only if usePrivateAgent is set in the options. We also need to introduce a destroy function on HttpClient so that consumers can call it to close connections on private agent. We decided to share the agent across requests only if usePrivateAgent is true, since different agent might be needed for different get calls (e.g one request may bypass proxy and other might need to use tunnel agent). So only if consumers are sure that they want to use same agent across requests they can set usePrivateAgent to true. I have also added options of keepAlive and maxSockets to configure the agent.

lib/HttpClient.ts
```diff
@@ -88,7 +88,8 @@ export class HttpClient {
     private _httpProxyBypassHosts: RegExp[];
     private _allowRedirects: boolean = true;
     private _maxRedirects: number = 50
-
+    private _agent;
+    private _usePrivateAgent;
```
bryanmacfarlane: this is no longer "private" agent.  it's whether we reuse connections.  rename

lib/HttpClient.ts
```diff
@@ -134,6 +135,8 @@ export class HttpClient {
             if (requestOptions.maxRedirects != null) {
                 this._maxRedirects = Math.max(requestOptions.maxRedirects, 0);
             }
+
+            this._usePrivateAgent = !!this.requestOptions.usePrivateAgent;
```
bryanmacfarlane: at first I thought, well let's just always re-use the agent for perf.  When wouldn't you want to.  The argument was brought that some calls are to proxy, some not etc...  But, constructor of this http client takes proxy config.  So if you wanted to make some calls to proxy, some not, you already need to contruct two instances.

However, I do acknowledge that there might be a case where you don't want to re-use within a single instance of the http client.  I can't come up with it but it might exist. 

That's what brought me to the conclusion that we should still have this option but it should be an option to *not* reuse agent. 

Other option I'm toying with is whether we have a proxy config with an exclusion list but maybe that's a higher level concept?  I need to look at .net http clients for patterns.  Will follow up.

bryanmacfarlane: I think we should follow the .net pattern of taking a bypass list of urls for proxy config.

https://msdn.microsoft.com/en-us/library/ms144293(v=vs.110).aspx

If we do that, then there is one proxy config.  That means the request code here should reuse an agent for non proxy calls and potentially one more if a proxy call.  

So, I think this should morph into "supporting proxy exclusion lists" and the agent should pass through since it supports exclusion lists.

lib/HttpClient.ts
omeshp: done

lib/HttpClient.ts
omeshp: That seems fine to me, I have updated the PR to have two agents one for proxy and another one for non-proxy and use the correct one based on the bypass list.

lib/HttpClient.ts
```diff
@@ -259,96 +273,136 @@ export class HttpClient {
         let usingSsl = info.parsedUrl.protocol === 'https:';
         info.httpModule = usingSsl ? https : http;
         var defaultPort: number = usingSsl ? 443 : 80;
+        info.options = <http.RequestOptions>{};
```
ericsciple: It would be great if this PR proactively had contextual clues for the reader. It looks like at least some of the diff is related to code moving around, which is hard to interpret.

omeshp: I agree unfortunately github PR experience isn't great to review such changes, I have tried to add comments where I though they would help. Mostly the change is in the _getAgent function which is used to get agent, rest of the info properties remain the same. In _getAgent we are either returning one of the private agents (_agent or _proxyAgent) depending on reuse or creating a new agent. If you need more info about any particular change I can help.

lib/HttpClient.ts
omeshp: done

lib/Interfaces.ts
```diff
@@ -22,7 +22,10 @@ export interface IRequestOptions {
     proxy?: IProxyConfiguration,
     cert?: ICertConfiguration,
     allowRedirects?: boolean, 
-    maxRedirects?: number
+    maxRedirects?: number,
+    reuseAgent? : boolean,
+    maxSockets?: number,
+    keepAlive?: boolean
```
bryanmacfarlane: After talking here, I think we just want a keepAlive option and if true, we reuse agents / connections.  There's no scenarios where I want keepAlive but not reuse.

Since keepAlive defaults to false in node let's default to false.

If you're using the rest client, I think this means you need to add that option to the rest client and it passes it through.

lib/HttpClient.ts
```diff
@@ -202,6 +206,16 @@ export class HttpClient {
         return response;
     }
 
+    public destroy() {
```
bryanmacfarlane: Let's call this dispose.  Put a comment on the method that it only needs to be called if keepAlive is true.  This should also set a disposed flag and other operations should fail if disposed.

lib/HttpClient.ts
```diff
@@ -259,96 +273,136 @@ export class HttpClient {
...
-        info.options.host = info.parsedUrl.hostname;
-        info.options.port = info.parsedUrl.port ? parseInt(info.parsedUrl.port) : defaultPort;
-        info.options.path = (info.parsedUrl.pathname || '') + (info.parsedUrl.search || '');
-        info.options.method = method;
-        info.options.headers = headers || {};
-        info.options.headers["User-Agent"] = this.userAgent;
+        var parsedUrl = url.parse(requestUrl);
+        let usingSsl = parsedUrl.protocol === 'https:';
+        let keepAlive = false;
+        let maxSockets = http.globalAgent.maxSockets;
```
bryanmacfarlane: Let's get rid of the http.globalAgent.maxSockets since you added the option.  Use the option

lib/Interfaces.ts
omeshp: done using keepAlive now. We aren't using rest client, but it looks like rest client is sharing the same interface (https://github.com/Microsoft/typed-rest-client/blob/master/lib/RestClient.ts#L37) so these options should work for rest client too.

lib/HttpClient.ts
omeshp: http.globalAgent.maxSockets is the default value, if value in options is specified it will be overridden below

lib/HttpClient.ts
omeshp: done.

bryanmacfarlane:
One more thing.  Add some unit tests to make sure it basically works.  Perhaps dupe some of the very basic crud http / rest tests with keepAlive true.  Just so we don't regress ...

lib/HttpClient.ts
```diff
@@ -88,7 +88,10 @@ export class HttpClient {
     private _httpProxyBypassHosts: RegExp[];
     private _allowRedirects: boolean = true;
     private _maxRedirects: number = 50
-
+    private _agent;
```
bryanmacfarlane: kill line 11 hard coded max.  instead, just default your options to 100 if not set (or should we not set)?  never liked the hard code setting which we never removed.  since you're added maxConnections options, let's take it out.

omeshp: default of 100 seems reasonable to me, will update PR.

Issue #31, opened 2017-12-04
add npm run buildtest (make.js) 


bryanmacfarlane:

bryanmacfarlane:
https://github.com/Microsoft/typed-rest-client/commit/085ce6acec4cfc4f305980457ec14c761d4b2f08

PR #35, opened 2017-12-04
Add npm run bt and npm run bts

stephenmichaelf:
Add npm run commands for:
- build and test (bt)
- build, test, and samples (bts).

I did them as abbreviated, seems easier to me. If you prefer the long way(buildtest, buildtestsamples) let me know! Or we could have both versions for all cases and add npm run b and npm run t.

#31


bryanmacfarlane:
oops - I didn't see this and pushed after doing this with other changes.  sorry

Issue #27, opened 2017-11-30
enforce node and npm version in make.js


bryanmacfarlane:
LTS Node >= 6.12.0 and Npm 5.5.1

stephenmichaelf:
https://github.com/Microsoft/vsts-tasks/blob/master/make.js#L59

PR #36, opened 2017-12-05
Enforce node and npm versions in make.js

stephenmichaelf:
- Requirement for node version
- Requirement for npm version
- Remove calls to build/clean from other methods. They need to be called on their own.
- Increase timeout for slow tests(run on my home machine)

#27

### DIFF
--- a/make.js
+++ b/make.js
@@ -17,7 +17,7 @@ var buildPath = path.join(__dirname, '_build');
 var testPath = path.join(__dirname, 'test');
 
 // enforce minimum Node version
-var minimumNodeVersion = '6.12.0';
+var minimumNodeVersion = '4.8.6';
 var currentNodeVersion = process.versions.node;
 if (semver.lt(currentNodeVersion, minimumNodeVersion)) {
     fail('requires node >= ' + minimumNodeVersion + '.  installed: ' + currentNodeVersion);
--- a/make.js
+++ b/make.js
@@ -42,8 +42,6 @@ target.clean = function () {
 };
 
 target.build = function () {
-    target.clean();
-
     run(path.join(__dirname, 'node_modules/.bin/tsc') + ' --outDir ' + buildPath);
     cp('-Rf', rp('lib/opensource'), buildPath);
     cp(rp('package.json'), buildPath);
@@ -52,8 +50,6 @@ target.build = function () {
 }
 
 target.test = function() {
-    target.build();
-
     // install the just built lib into the test proj
     pushd('test')
     run('npm install ../_build');
@@ -64,8 +60,6 @@ target.test = function() {
 }
 
 target.samples = function () {
-    target.build();
-
     pushd('samples');
     run('node samples.js');
     popd();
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typed-rest-client",
-  "version": "0.16.0",
+  "version": "1.0.0",
```
stephenmichaelf: Not sure why this happened.

bryanmacfarlane: note that 1.0.0 is released with a preview tag.  So, you only get if you do npm install typed-rest-client@preview.

make.js
```diff
@@ -9,6 +11,21 @@ var rp = function (relPath) {
 var buildPath = path.join(__dirname, '_build');
 var testPath = path.join(__dirname, 'test');
 
+// enforce minimum Node version
+var minimumNodeVersion = '6.12.0';
+var currentNodeVersion = process.versions.node;
+if (semver.lt(currentNodeVersion, minimumNodeVersion)) {
+    fail('requires node >= ' + minimumNodeVersion + '.  installed: ' + currentNodeVersion);
```
bryanmacfarlane: where is fail defined?  did I miss it?

stephenmichaelf: Thought I had pushed, it's there now.

make.js
```diff
@@ -1,14 +1,36 @@
...
+var fail = function (message) {
+    console.error('ERROR: ' + message);
+    process.exit(1);
+}
+
 var buildPath = path.join(__dirname, '_build');
 var testPath = path.join(__dirname, 'test');
 
+// enforce minimum Node version
+var minimumNodeVersion = '6.12.0';
```
bryanmacfarlane: I tested and this works with lts/argon -> v4.8.6.  lets make that the minimum.  Since we have to test that it runs on argon, it's convenient to dev on it (use nvm and then build, test, samples.  Alternatively, we could keep 6.12 as dev minimum but then to test on argon you would have to build on 6.12 and then nvm to 4.x and then run tests.

bryanmacfarlane:
other than note about 4.8.6, looks good.  resolve conflicts and then I'll squash and merge

make.js
stephenmichaelf: It seems worthwhile to update our documentation to say what node versions you can run against. 

I was also thinking of adding a make.js step for that automates what you just said(using nvm), building and then running tests across all supported node versions. Of course this assumes that nvm is installed and setup correctly(we could also add a courtesy nvm install for missing versions) but it seems useful to have this ability during development to say run allversiontests (or something along these lines) and know you are getting full testing before pushing.

What do you think?

bryanmacfarlane: Yeah, I was thinking the same on both fronts.  Go ahead.

stephenmichaelf: Great, I created new tickets for to track them.

stephenmichaelf:
Node version lowered and master merged.

One thing I realized... it can be strange to use the lower node versions to build since you need the higher npm version. I don't think nvm handles this?

When we update the documentation I plan to put that there is a suggested version to build with (8.9.1) so that the UX is easy for using nvm, you would have the right npm version too. Otherwise if people want to build with a lower node version they would have to change the node version using nvm then change the npm version to be 5.5.1.

PR #40, opened 2017-12-07
Compile samples before running

djKooks:
I added `tsc` to compile ts file in /sample before testing it, because `sample.js` cannot be found before compile.

Please look on. Thanks.

### DIFF
--- a/make.js
+++ b/make.js
@@ -48,7 +48,7 @@ target.samples = function () {
     target.build();
 
     pushd('samples');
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
Thanks for taking the time to create a PR.  I had already fixed in local changes of mine.  Fixed here https://github.com/Microsoft/typed-rest-client/commit/68d80682d313817f4db185464686896119b9f5c1

Issue #39, opened 2017-12-06
Linting and strict compilation


stephenmichaelf:
Look at:
- tslint
- tslint-microsoft-contrib
- strict compile settings

stephenmichaelf:
This will be added but we are deferring the work at the moment to focus on higher priorities.

nschonni:
Probably fine to do TSLint for now, but it looks like ESLint may replace it in the future https://eslint.org/blog/2019/01/future-typescript-eslint

PR #43, opened 2017-12-07
Add Linting

stephenmichaelf:
Are you OK with going down this road? It seems worthwhile to me for the consistency, quality, and safety. It will take time to do the initial work but then each PR worth of changes should be small.

They suggest LF line endings too, I changed those. Seems like GitHub may be ignoring that.

Some things that have been caught:
- Missing  semicolons
- Cyclomatic complexity
- Formatting
- Unresolved promise
- Unnecessary variables

My preference would be to do PRs every day or so and comment out the lint method call until its done. There are still a large number of things to fix but we need the code to build obviously.

This way we dont get one huge merge that is hard to review.

What do you think? Then we get in line with the security practices and code standards.

If you agree let me know and I can let you know when this is ready to merge for the day.

#39

### DIFF
--- a/lib/Handlers.ts
+++ b/lib/Handlers.ts
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
 export { BasicCredentialHandler } from "./handlers/basiccreds";
 export { BearerCredentialHandler } from "./handlers/bearertoken";
 export { NtlmCredentialHandler } from "./handlers/ntlm";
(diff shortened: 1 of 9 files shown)


bryanmacfarlane:
I'm fine but Nyoman is higher pri

Get Outlook for iOS<https://aka.ms/o0ukef>

stephenmichaelf:
What does Nyoman mean? :)

bryanmacfarlane:
According to my phone type ahead, it means ntlm

stephenmichaelf:
Let's discuss that when you can, there are more changes there than I thought.

This is ready for merge after you review.

One thing I am unsure of: in package-lock.json some of the integrity items changed to sha1-5, not sure why.

bryanmacfarlane:
I like catching real issues (promise chains etc...) but lets be careful about high churn for low value rules (variable names etc...)

stephenmichaelf:
I am going to get rid of this PR and we can start from a simple linting setup with high value options turned on. This one is too heavy handed with low ROI on some things.

PR #46, opened 2017-12-08
Fix issue with destroying proxy agent

omeshp:
Proxy agent doesn't support destroy function

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -218,11 +218,7 @@ export class HttpClient {
         if (this._agent) {
             this._agent.destroy();
         }
-
-        if (this._proxyAgent) {
-            this._proxyAgent.destroy();
-        }
-
+        
         this._disposed = true;
     }
 


lib/HttpClient.ts
```diff
@@ -218,11 +218,7 @@ export class HttpClient {
         if (this._agent) {
             this._agent.destroy();
         }
-
-        if (this._proxyAgent) {
-            this._proxyAgent.destroy();
-        }
-
+
```
bryanmacfarlane: what about keep alives?  Does proxy support that?

bryanmacfarlane:
@TingluoHuang for review and proxy testing

lib/HttpClient.ts
bryanmacfarlane: ??

omeshp: From the testing I didn't it seemed to reuse sockets, but at https://github.com/koichik/node-tunnel/blob/master/lib/tunnel.js I don't see any keepAlive. I will have to retest to validate. I can try it out on Friday.

omeshp: Ahh it is indeed getting assigned at https://github.com/koichik/node-tunnel/blob/master/lib/tunnel.js#L47

bryanmacfarlane: I also asked Ting to test and chime in on Proxy since he did most of that work.

TingluoHuang: https://github.com/Microsoft/typed-rest-client/blob/master/lib/HttpClient.ts#L327
if i don't provide requestOptions.maxSockets, then Keep-Alive won't work.
i would do`maxSockets = this.requestOptions.maxSockets || 100` instead of `maxSockets = this.requestOptions.maxSockets || http.globalAgent.maxSockets` since ` http.globalAgent.maxSockets` is `int.max`

TingluoHuang: why no longer need to destroy proxyagent?

stephenmichaelf: Ting could this be causing my keep alive to fail? The lack of max sockets

omeshp: @TingluoHuang keepalive will still work if maxSocket is Infinity. maxSockets is for throttling requests. In node agent (https://nodejs.org/api/http.html#http_class_http_agent) the default value of maxSockets is Infinity, but I am ok with restricting it to 100 here.

proxyAgent doesn't support destroy function, so we don't have a choice.

TingluoHuang: @stephenmichaelf it's possible if you are using proxy at the same time.
@omeshp i think there might be something wrong in the tunnel lib which cause the problem.

omeshp: @TingluoHuang I didn't get the issue stephen is facing. Are you saying that keepAlive isn't honored in case of proxy if maxSockets isn't set?

stephenmichaelf:
@TingluoHuang @omeshp Where are we on this?

lib/HttpClient.ts
omeshp: @TingluoHuang @stephenmichaelf I think tunnel agent maintains list of sockets and reuses them (https://github.com/koichik/node-tunnel/blob/master/lib/tunnel.js#L58) by default, so sending keepAlive in proxy case won't matter, but this PR should go through since destroy function isn't supported in tunnel agent and users will get error while calling dispose on typed-rest-client.

stephenmichaelf: @TingluoHuang Maybe this is why we were seeing unexpected keep-alive results? Since this code is using tunnel and not the standard node sockets.

@omeshp To clarify, this is different than node's standard usage of sockets. We had some issues adding NTLM to this library due to node dropping sockets if there wasn't a request queued up on the socket when the initial request finished. I think this unique functionality in tunnel to reuse sockets is why this code is working. In NTLM we ended up using callbacks and setImmediate. Are you familiar with this? Is this what you would expect?

bryanmacfarlane:
released as 1.0.6

Issue #49, opened 2017-12-20
Proxy doesn't work when password has url sensitive characters


omeshp:
### Environment
Node version: 6.x
Npm version: 
OS and version: 
typed-rest-client version: 0.15.1

### Issue Description
If password has characters like ':'  '\'  proxy doesn't work, since we are url encoding the password at 
https://github.com/Microsoft/typed-rest-client/blob/93448b8ffdf5183cddb710d0fde99d7e697026f1/lib/HttpClient.ts#L408

### Expected behaviour
Proxy should work when password has url sensitive characters.

### Actual behaviour
Error: tunneling socket could not be established, statusCode=407

TingluoHuang:
@omeshp we need `encodeURIComponent` because we want to allow proxy password contains reserved chars. `;,/?:@&=+$`

omeshp:
@TingluoHuang password is sent as a part of header so they shouldn't be URL encoded. Currently reserved characters aren't working with proxy, this fix is to make them work.

TingluoHuang:
@omeshp Got it. Thanks for finding and fixing this. Changes looks good to me.

PR #50, opened 2017-12-20
Fix proxy if password has url sensitive characters

omeshp:


omeshp:
#49

TingluoHuang:
what will happen when my password contains any of these `;,/?:@&=+$`, like `http://user:pass?word@github.com`

omeshp:
@TingluoHuang tunnel library doesn't send password as a part of URL, it's set in header.

TingluoHuang:
Looks good to me.  thanks for finding and fixing this!

bryanmacfarlane:
+1 - I can release after you merge

Issue #42, opened 2017-12-07
Test all versions using make.js


stephenmichaelf:
Automate building and testing against all supported node versions using nvm.

Potentially call it testallversions.

stephenmichaelf:
Discussed offline, we are not going to add this feature.

PR #52, opened 2018-01-04
Test all supported versions of node

stephenmichaelf:
#42 

This change adds a way for us to test all supported node versions locally.

### DIFF
--- a/make.js
+++ b/make.js
@@ -17,17 +17,6 @@ var runCommand = function (command) {
     return ncp.execSync(command, { encoding: 'utf-8' });
 }
 
-var getCurrentNodeVersionFromNvm = function () {
-    var nvmList = runCommand('nvm list');
-
-    // find the row that starts with *, that is the current version
-
-    // it may say "(Currently using 64-bit executable)", strip that out
-
-
-    return '';
-}
-
 var buildPath = path.join(__dirname, '_build');
 var testPath = path.join(__dirname, 'test');
 
@@ -86,6 +75,7 @@ target.test = function() {
     run('mocha test');
 }
 
+// TODO: Add notes to documentation that this needs to run as admin. And how to use it.
 target.testall = function() {
     // make sure we have nvm
     var nvmList = runCommand('nvm');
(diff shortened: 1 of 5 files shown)


bryanmacfarlane:
Works fine on windows?  I didn't think nvm worked on windows??

stephenmichaelf:
Yes sir, works on Windows. I use [this one](https://github.com/coreybutler/nvm-windows) which I believe functions the same as the non-windows version. Did the tests work on Mac?

make.js
```diff
@@ -71,6 +75,77 @@ target.test = function() {
...
 }
 
+target.testall = function() {
+    // make sure we have nvm
+    var nvmCommand = runCommand('nvm');
+    if (nvmCommand.indexOf('Running version') === -1) {
+        fail('requires nvm to be installed');
+    }
+    
+    var versionsToTest = ['8.9.1', '6.12.0', '4.8.6'];
```
bryanmacfarlane: This is already out of date.  6.12.3 is latest LTS.
Why not simplify all this logic here and simply use, 4, 6, 8
Then run nvm install x and nvm use x everytime.

The only time nvm install will take a time hit is when there's a new version and that's what we want.

stephenmichaelf: According to [this](https://stackoverflow.com/questions/44789330/use-nvm-to-install-latest-nodejs-for-specific-major-version) we should be able to say "nvm install 8" and get the latest LTS but it's not working, at least for the Windows version. I am guessing there are differences in the versions but not sure.

One alternative would be to do "nvm list available" and parse that, but it seems like overkill. We could also make a REST call I am guessing but that too seems like overkill. I will keep looking into this.

stephenmichaelf: After looking [here](https://github.com/coreybutler/nvm-windows/issues/156) it looks like the Windows version behaves differently, I requested a change but looks like we will need a workaround that works cross plat.

stephenmichaelf: Discussed offline, plan is to make it work for Linux/OSX and the command will do nothing on Windows.

stephenmichaelf:
Discussed offline, we are removing this feature.

Issue #16, opened 2017-11-17
Add back NTLM


stephenmichaelf:

bryanmacfarlane:
Support was lost in latest version ...

peterblazejewicz:
Do you mean adding it back via NPM? #22

bryanmacfarlane:
No, I mean in the current version of this library ntlm support is not wired up because we switched to async await and we needed to convert some of the callback code to promises.  There's no artificial constraint that it's done via an npm package which also carries an unneeded and couple http library.

bryanmacfarlane:
More info - httpclient used to be part of node api.  ntlm was added here

https://github.com/Microsoft/vsts-node-api/pull/43/files#diff-a218a7865550fc62c59bb408113064f3

We broke http client out so it can be used in many of our other libs like our tool lib.  When we broke it out, we made it fully async await and ditched call backs.  There was a TODO into this code to re-introduce it but with promises/async/await.   So the handler signature needs to change.  It's one of the things that we're waiting for to release this.  We want that interface to settle.

bryanmacfarlane:
ready for review?

stephenmichaelf:
Not yet. My fix didn't work. I found any other network traffic analyzer and I'm going to try and compare the working vs not working versions. The NTLM strings seem to be generated fine and the steps are all happening. I think it's an issue with adding/removing headers(joining objects incorrectly) or with reusing the connection. Continuing to dig.

stephenmichaelf:
There is a very ugly branch pushed up with a bunch of logging and stuff that I've been testing on.

bryanmacfarlane:
Done and released in latest 1.0.6

PR #56, opened 2018-01-11
Add NTLM authorization handler.

stephenmichaelf:
NOTE: I combined all my changes into one commit to make things cleaner.

This change adds an authorization handler for NTLM.

Due to the way NTLM works on connections there is some refactoring to expose a raw request in the HttpClient that uses callbacks. The method that existed before, requestRaw, calls in to this. There were some bugs regarding hanging sockets so that is why there is code to readBody on the responses before making the next request.

#16

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -254,7 +254,7 @@ export class HttpClient implements ifm.IHttpClient {
      */
     public requestRaw(info: ifm.IRequestInfo, data: string | NodeJS.ReadableStream): Promise<ifm.IHttpClientResponse> {
         return new Promise<ifm.IHttpClientResponse>((resolve, reject) => {
-            var callbackForResult = function (err: any, res: ifm.IHttpClientResponse) {
+            let callbackForResult = function (err: any, res: ifm.IHttpClientResponse) {
                 if (err) {
                     reject(err);
                 }
@@ -280,8 +280,8 @@ export class HttpClient implements ifm.IHttpClient {
             info.options.headers["Content-Length"] = Buffer.byteLength(data, 'utf8');
         }
 
-        var callbackCalled: boolean = false;
-        var handleResult = (err: any, res: HttpClientResponse) => {
+        let callbackCalled: boolean = false;
+        let handleResult = (err: any, res: HttpClientResponse) => {
             if (!callbackCalled) {
                 callbackCalled = true;
                 onResult(err, res);
@@ -371,15 +371,15 @@ export class HttpClient implements ifm.IHttpClient {
             return agent;
         }
 
-        var parsedUrl = url.parse(requestUrl);
-        let usingSsl = parsedUrl.protocol === 'https:';
+        let parsedUrl = url.parse(requestUrl);
+        const usingSsl = parsedUrl.protocol === 'https:';
         let maxSockets = 100;
         if (!!this.requestOptions) {
             maxSockets = this.requestOptions.maxSockets || http.globalAgent.maxSockets
         }
 
         if (useProxy) {
-            var agentOptions: tunnel.TunnelOptions = {
+            const agentOptions: tunnel.TunnelOptions = {
                 maxSockets: maxSockets,
                 keepAlive: this._keepAlive,
                 proxy: {
@@ -389,8 +389,8 @@ export class HttpClient implements ifm.IHttpClient {
                 },
             };
 
-            var tunnelAgent: Function;
-            var overHttps = proxy.proxyUrl.protocol === 'https:';
+            let tunnelAgent: Function;
+            const overHttps = proxy.proxyUrl.protocol === 'https:';
             if (usingSsl) {
                 tunnelAgent = overHttps ? tunnel.httpsOverHttps : tunnel.httpsOverHttp;
             } else {
@@ -403,7 +403,7 @@ export class HttpClient implements ifm.IHttpClient {
 
         // if reusing agent across request and tunneling agent isn't assigned create a new agent
         if (this._keepAlive && !agent) {
-            var options = { keepAlive: this._keepAlive, maxSockets: maxSockets };
+            const options = { keepAlive: this._keepAlive, maxSockets: maxSockets };
             agent = usingSsl ? new https.Agent(options) : new http.Agent(options);
             this._agent = agent;
         }
@@ -428,7 +428,7 @@ export class HttpClient implements ifm.IHttpClient {
     }
 
     private _getProxy(requestUrl) {
-        var parsedUrl = url.parse(requestUrl);
+        const parsedUrl = url.parse(requestUrl);
         let usingSsl = parsedUrl.protocol === 'https:';
         let proxyConfig: ifm.IProxyConfiguration = this._httpProxy;
 
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -36,9 +36,9 @@ export enum HttpCodes {
     GatewayTimeout = 504,
 }
 
-const HttpRedirectCodes: number[] = [ HttpCodes.MovedPermanently, HttpCodes.ResourceMoved, HttpCodes.TemporaryRedirect, HttpCodes.PermanentRedirect ];
+const HttpRedirectCodes: number[] = [HttpCodes.MovedPermanently, HttpCodes.ResourceMoved, HttpCodes.TemporaryRedirect, HttpCodes.PermanentRedirect];
 
-export class HttpClientResponse {
+export class HttpClientResponse implements ifm.IHttpClientResponse {
     constructor(message: http.IncomingMessage) {
         this.message = message;
     }
@@ -71,11 +71,11 @@ export function isHttps(requestUrl: string) {
 }
 
 enum EnvironmentVariables {
-    HTTP_PROXY = "HTTP_PROXY", 
+    HTTP_PROXY = "HTTP_PROXY",
     HTTPS_PROXY = "HTTPS_PROXY",
 }
 
-export class HttpClient {
+export class HttpClient implements ifm.IHttpClient {
     userAgent: string;
     handlers: ifm.IRequestHandler[];
     requestOptions: ifm.IRequestOptions;
@@ -97,7 +97,7 @@ export class HttpClient {
 
     constructor(userAgent: string, handlers?: ifm.IRequestHandler[], requestOptions?: ifm.IRequestOptions) {
         this.userAgent = userAgent;
-        this.handlers = handlers;
+        this.handlers = handlers || [];
         this.requestOptions = requestOptions;
         if (requestOptions) {
             if (requestOptions.ignoreSslError != null) {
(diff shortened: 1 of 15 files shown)


lib/HttpClient.ts
```diff
@@ -142,35 +142,35 @@ export class HttpClient {
         }
     }
 
-    public options(requestUrl: string, additionalHeaders?: ifm.IHeaders): Promise<HttpClientResponse> {
+    public options(requestUrl: string, additionalHeaders?: ifm.IHeaders): Promise<ifm.IHttpClientResponse> {
```
ericsciple: does this break consumers on a patch update?

lib/HttpClient.ts
```diff
@@ -226,63 +247,93 @@ export class HttpClient {
...
-                let res: HttpClientResponse = new HttpClientResponse(msg);
-                resolve(res);
-            });
+    /**
+     * Raw request with callback.
+     * @param info 
+     * @param data 
+     * @param onResult 
+     */
+    public requestRawWithCallback(info: ifm.IRequestInfo, data: string | NodeJS.ReadableStream, onResult: (err: any, res: ifm.IHttpClientResponse) => void): void {
```
ericsciple: have you considered an actual overload rather than two functions `requestRaw` and `requestRawWithCallback`?

ericsciple: or is the return type different, is that the point?

ericsciple: do these both need to be public? do we want to support both interfaces?

ericsciple: (also, if i remember correctly, i think overloads can have different return types so even if we want to support both as public i think it could simply be an overload. i think that would be a more common API interface, unless these function names were chosen to match some other library's pattern (e.g. node)?

lib/HttpClient.ts
```diff
@@ -226,63 +247,93 @@ export class HttpClient {
...
-                reject(err);
-            });
+        req.on('socket', (sock) => {
+            socket = sock;
+        });
 
-            if (data && typeof (data) === 'string') {
-                req.write(data, 'utf8');
+        // If we ever get disconnected, we want the socket to timeout eventually
+        req.setTimeout(this._socketTimeout || 3 * 60000, () => {
```
ericsciple: is this 3 minutes? can you add a comment

lib/HttpClient.ts
```diff
@@ -226,63 +247,93 @@ export class HttpClient {
...
-                resolve(res);
-            });
+    /**
+     * Raw request with callback.
+     * @param info 
+     * @param data 
+     * @param onResult 
+     */
+    public requestRawWithCallback(info: ifm.IRequestInfo, data: string | NodeJS.ReadableStream, onResult: (err: any, res: ifm.IHttpClientResponse) => void): void {
+        let socket;
```
ericsciple: i see you explicitly defined `callbackCalled` and `handleResult` using `var` (because they are using in the callbacks?), yet `socket` using `let`. is this intentional?

lib/HttpClient.ts
ericsciple: i now see this is largely existing code with the addition of handleResult

what problem existed that you are solving with handleResult? reject called twice? would be great to have some pr comments for the reader explaining this. it takes a while to attempt to infer from the diff, especially since this method is largely syntax changes and the diff is hard to read.

ericsciple: what was the symptom before? did we have a case where it manifested in strange behavior? or did you just notice a bug?

lib/handlers/personalaccesstoken.ts
```diff
@@ -18,10 +18,11 @@ export class PersonalAccessTokenCredentialHandler implements ifm.IRequestHandler
     }
 
     // This handler cannot handle 401
-    canHandleAuthentication(res: ifm.IHttpResponse): boolean {
+    canHandleAuthentication(response: ifm.IHttpClientResponse): boolean {
         return false;
     }
 
-    handleAuthentication(httpClient, protocol, options, objs, finalCallback): void {
+    handleAuthentication(httpClient: ifm.IHttpClient, requestInfo: ifm.IRequestInfo, objs): Promise<ifm.IHttpClientResponse> {
```
ericsciple: lots of interface changes. is this because it's still in preview?

lib/HttpClient.ts
stephenmichaelf: I tried to do an overload and the compiler was not happy with it, I can look into it more deeply. We may not need the non-callback version to be public but we need the callback version to be public for the authorization workflow. It has an HttpClient that it needs to call this method on in order to make the NTLM requests.

Yes return type and params are different, one returns the response and takes requestinfo and data and one is void but takes in a callback and input is requestinfo, data, and a callback. I looked up a bit on the overloading and it seems like you can only declare the overload but not implement it, there can only be one implementation. For instance add(string foo); add(int foo) {}. But only the second has an implementation and you check typeof foo. I could be misunderstanding it.

lib/HttpClient.ts
stephenmichaelf: It shouldn't, I added an interface but the structure is the same as it was before(HttpClientResponse).

lib/handlers/personalaccesstoken.ts
stephenmichaelf: Yes. This is a symptom of this code having been copied from another repo then modified. Also for these handlers since NTLM is the only one that can handleAuthentication and it wasn't being used before I don't think we break anything with this specific change. PAT uses prepareRequest which didn't change signature.

lib/HttpClient.ts
stephenmichaelf: Unintentional, will fix.

lib/HttpClient.ts
ericsciple: Once the implementation adds something that the interface doesn't have, it will break consumers. Consider this:

image

bryanmacfarlane: I think this is fine because the new major version (1.x) and all it's patches have been in preview.

lib/HttpClient.ts
stephenmichaelf: Will add a comment for the default timeout.

stephenmichaelf: The formatting of the diff got really messed up, it looks terrible.

This change is due to adding NTLM. The way keep alive connections work is that a subsequent request has to be queued on the socket when the first request finishes otherwise the socket closes. When using async/await I kept getting an HTTP 401 Unauthorized. My thought was that the way I was using setImmediate in conjunction with async/await caused the second request to happen too late in the node event loop so the socket was closing. For NTLM if the request is on a different connection then it fails, both NTLM messages have to be on the same connection.

In order to fix this I refactored part of the code to use callbacks so that I could call setImmediate inside the callback of the first request. I believe this changes when in the node event loop the request gets queued and ensures that the second request is pushed to the connection request queue before the nextTick in the event loop(this must be when the connection checks if anything is queued and drops the connection if the request queue is empty). This allows the requests to both happen on the same connection and the NTLM handshake to be successful.

I will note that it's possible I wasn't doing setImmediate in conjunction with async/await correctly but I messed around with it for a while and couldn't get it working and the other code we copied from to do NTLM all uses callbacks. I was seeing some strangeness with readBody on the response message that also caused a 401 so it's possible I was conflating the two issues and if ere readBody is done correctly and the code uses async/await  it works but I am not sure. If it makes sense I can try to follow this path and see. Unfortunately testing NTLM is very time consuming and tricky to debug.

lib/HttpClient.ts
```diff
@@ -226,63 +247,93 @@ export class HttpClient {
...
 
-    private _requestRaw(info: RequestInfo, data: string | NodeJS.ReadableStream): Promise<HttpClientResponse> {
-        return new Promise<HttpClientResponse>((resolve, reject) => {
-            let socket;
+    /**
+     * Raw request.
+     * @param info 
+     * @param data 
+     */
+    public requestRaw(info: ifm.IRequestInfo, data: string | NodeJS.ReadableStream): Promise<ifm.IHttpClientResponse> {
```
ericsciple: do we need to expose these publicly? can't users turn off redirects and not pass auth handlers to get the same behavior? extra complexity to have multiple ways to achieve the same goal.

lib/handlers/ntlm.ts
```diff
@@ -2,124 +2,160 @@
...
             // Once we have the www-authenticate header, split it so we can ensure we can talk NTLM
-            var wwwAuthenticate = res.headers['www-authenticate'];
-            if (wwwAuthenticate !== undefined) {
-                var mechanisms = wwwAuthenticate.split(', ');
-                var idx =  mechanisms.indexOf("NTLM");
-                if (idx >= 0) {
+            const wwwAuthenticate = response.message.headers['www-authenticate'];
+
+            if (wwwAuthenticate) {
+                const mechanisms = wwwAuthenticate.split(', ');
```
ericsciple: is this safe or should we split on comma and then trim and if not empty, then proceed

lib/handlers/ntlm.ts
```diff
@@ -2,124 +2,160 @@
...
+                const index = mechanisms.indexOf("NTLM");
+                if (index >= 0) {
                     // Check specifically for 'NTLM' since www-authenticate header can also contain
                     // the Authorization value to use in the form of 'NTLM TlRMTVNT....AAAADw=='
-                    if (mechanisms[idx].length == 4) {
+                    if (mechanisms[index].length == 4) {
```
ericsciple: length == 4 is redundant

lib/handlers/ntlm.ts
```diff
@@ -2,124 +2,160 @@
...
+
         let self = this;
+
         // The following pattern of sending the type1 message following immediately (in a setImmediate) is
         // critical for the NTLM exchange to happen.  If we removed setImmediate (or call in a different manner)
         // the NTLM exchange will always fail with a 401.
-        this.sendType1Message(httpClient, protocol, ntlmOptions, objs, keepaliveAgent, function (err, res) {
+        this.sendType1Message(httpClient, requestInfo, objs, function (err, res) {
             if (err) {
                 return finalCallback(err, null, null);
```
ericsciple: this never rejects

ericsciple: note, in the callback, can readbody can have an error too?

lib/handlers/ntlm.ts
```diff
@@ -2,124 +2,160 @@
...
             }
-            setImmediate(function () {
-                self.sendType3Message(httpClient, protocol, ntlmOptions, objs, keepaliveAgent, res, finalCallback);
+
+            /// We have to readbody on the response before continuing otherwise there is a hang.
+            res.readBody().then(() => { 
+                setImmediate(function () {
```
ericsciple: consider adding a comment about setimmediate and socket disconnect problem

lib/handlers/ntlm.ts
```diff
@@ -2,124 +2,160 @@
...
+
+        const type3options: http.RequestOptions = {
             headers: {
-                'Authorization': type3msg
+                'Authorization': type3msg,
+                'Connection': 'Close'
             },
-            allowRedirects: false,
-            agent: keepaliveAgent
+            //allowRedirects: false,
```
ericsciple: remove commented code

ericsciple:
makes sense

lib/HttpClient.ts
stephenmichaelf: Fixed.

lib/handlers/ntlm.ts
stephenmichaelf: Done.

lib/handlers/ntlm.ts
stephenmichaelf: Done.

lib/handlers/ntlm.ts
stephenmichaelf: Added a reject if(err). In the readbody code it never rejects either but I supposed theoretically it could? I would have to dig into http.IncomingMessage more. I didn't modify that code and we haven't seen any issues so I think it's OK as is as far as readBody goes.

lib/HttpClient.ts
stephenmichaelf: We do give them the request method which is relatively low level. We need requestRawWithCallbacks for the NTLM auth handler to use but we may not need this one. I would be OK putting this one back to private and only having the version with callbacks be public for the auth handlers and others can use it as their low level request if need be.

@bryanmacfarlane Does that work for you?

lib/handlers/ntlm.ts
stephenmichaelf: According to [the spec](https://tools.ietf.org/html/rfc7235#section-4.1) it looks like there will always be a space after commas.

lib/handlers/ntlm.ts
stephenmichaelf: Fixed.

stephenmichaelf:
Outside of the discussion on requestRaw/Callback visibility all the feedback is in.

@bryanmacfarlane @TingluoHuang any other feedback/concerns?

lib/HttpClient.ts
```diff
@@ -226,63 +247,93 @@ export class HttpClient {
...
+    /**
+     * Raw request with callback.
+     * @param info 
+     * @param data 
+     * @param onResult 
+     */
+    public requestRawWithCallback(info: ifm.IRequestInfo, data: string | NodeJS.ReadableStream, onResult: (err: any, res: ifm.IHttpClientResponse) => void): void {
+        let socket;
+        
+        let isDataString = typeof (data) === 'string';
```
TingluoHuang: this is not been used. :)

PR #57, opened 2018-01-16
Return 'null' on status 404

andrew8er:
This PR changes `RestClient._processResult()` to return a null result to match the behavior of the documentation. It previously returned `undefined`. It also updates the signaure of `RestResponse<T>` accordingly.

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -174,9 +174,12 @@ export class RestClient {
 
     private async _processResponse<T>(res: httpm.HttpClientResponse, options: IRequestOptions): Promise<IRestResponse<T>> {
         return new Promise<IRestResponse<T>>(async (resolve, reject) => {
-            let rres: IRestResponse<T> = <IRestResponse<T>>{};
-            let statusCode: number = res.message.statusCode;
-            rres.statusCode = statusCode;
+            const statusCode: number = res.message.statusCode;
+
+            const rres: IRestResponse<T> = {
+                statusCode: statusCode,
+                result: null,
+            };
 
             // not found leads to null obj returned
             if (statusCode == httpm.HttpCodes.NotFound) {
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -176,14 +176,14 @@ export class RestClient {
         return new Promise<IRestResponse<T>>(async (resolve, reject) => {
             const statusCode: number = res.message.statusCode;
 
-            const rres: IRestResponse<T> = {
+            const response: IRestResponse<T> = {
                 statusCode: statusCode,
                 result: null,
             };
 
             // not found leads to null obj returned
             if (statusCode == httpm.HttpCodes.NotFound) {
-                resolve(rres);
+                resolve(response);
             }
 
             let obj: any;
@@ -194,10 +194,10 @@ export class RestClient {
                 if (contents && contents.length > 0) {
                     obj = JSON.parse(contents);
                     if (options && options.responseProcessor) {
-                        rres.result = options.responseProcessor(obj);
+                        response.result = options.responseProcessor(obj);
                     }
                     else {
-                        rres.result = obj;
+                        response.result = obj;
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


lib/RestClient.ts
```diff
@@ -174,9 +174,12 @@ export class RestClient {
 
     private async _processResponse<T>(res: httpm.HttpClientResponse, options: IRequestOptions): Promise<IRestResponse<T>> {
         return new Promise<IRestResponse<T>>(async (resolve, reject) => {
-            let rres: IRestResponse<T> = <IRestResponse<T>>{};
-            let statusCode: number = res.message.statusCode;
-            rres.statusCode = statusCode;
+            const statusCode: number = res.message.statusCode;
+
+            const rres: IRestResponse<T> = {
```
stephenmichaelf: Can we rename this to response or something more readable?

test/resttests.ts
```diff
@@ -38,75 +38,75 @@ describe('Rest Tests', function () {
 
         let restRes: restm.IRestResponse<HttpBinData> = await _rest.get<HttpBinData>('https://httpbin.org/get');
         assert(restRes.statusCode == 200, "statusCode should be 200");
-        assert(restRes.result.url === 'https://httpbin.org/get');
+        assert(restRes.result && restRes.result.url === 'https://httpbin.org/get');
```
stephenmichaelf: For all of these, I am not sure it adds value to have the restRes.result check before doing an equality check. It fails either way and since this is just a test I don't think we need it. Makes it harder to read too.

test/resttests.ts
```diff
@@ -128,9 +128,9 @@ describe('Rest Tests', function () {
     it('gets a non-existant resource (404)', async() => {
         try {
             let restRes: restm.IRestResponse<HttpBinData> = await _rest.get<HttpBinData>('https://httpbin.org/status/404');
-            
+
             assert(restRes.statusCode == 404, "statusCode should be 404");
-            assert(restRes.result == null, "object should be null");
+            assert(restRes.result === null, "object should be null");
```
stephenmichaelf: result object should be null?

test/resttests.ts
stephenmichaelf: Same comment for all places this has been added.

andrew8er: I added these checks because I usually compile code with "strict" type checking enabled. If this option is disabled, then it would produces no error message, but in my eye, it is actually semantically wrong.

test/resttests.ts
andrew8er: According to your documentation in https://github.com/Microsoft/typed-rest-client/blob/master/README.md a 404 should return `null`.

andrew8er:
Actually I think the whole implementation of `_processResponse()` is rather convoluted. I have an refactored version for which I can provide a PR in the next days, but this would be a breaking change, as it alters the method's behavior a little bit.

In the current version, exceptions from the JSON parser and response processor get swallowed and a `null` response is generated. I think it makes more sense to let those exceptions bubble up to the caller.

lib/RestClient.ts
andrew8er: I retained the original variable name, but yes, this should be renamed.

bryanmacfarlane:
@andrew8er - I'm fine not holding compat on "internal" underscore methods.  A separate PR if you have a proposal for that.

test/resttests.ts
bryanmacfarlane: Yep - in the REST layer since it's job is to return a restful object, a 404 should return a null.  However the http layer will return body.

bryanmacfarlane: I think I'm fine with these lines / changes

stephenmichaelf: I did a poor job of writing that comment. I meant that the string could be changed to be more expressive.

stephenmichaelf:
@andrew8er Could you rename that variable at the top? Then we can get this merged. The null check on the tests should be fine. As @bryanmacfarlane mentioned we can do a separate PR for refactoring _processResponse(). Thanks again!

PR #58, opened 2018-01-30
Added HMAC-SHA256 Signature handler.

x51xxx:
Added  HMAC-SHA256 Signature handler.
Added request data in prepareRequest


bryanmacfarlane:
@x51xxx -  Thanks for the contribution.  Can you outline some context for the PR?  I'm also struggling between keeping it lightweight with other libs providing handlers (with readme pointing to other handlers) and offering value.  If common need will consider.

I guess either way, we need the stream option in prepare request, right?

Thoughts?

stephenmichaelf:
Bump @x51xxx

stephenmichaelf:
Going to close this for now due to inactivity. Please reopen with additional details when you get a chance. Thanks!

Issue #66, opened 2018-04-02
Empty base URL results in strange behavior with non-default ports


etiago:
### Environment
Node version: 8.10
Npm version: 5.8.0
OS and version: MacOS 10.13.3
typed-rest-client version: stable

### Issue Description
Using Typed Rest Client without specifying a base URL does not play well with non-default ports. I was using a node server on port 3000 and without specifying a URL, it tries to send requests as follows:
```
http://localhost:3000/:80/product/abca
```

### Expected behaviour
I expect it to behave like the `fetch` polyfill, where as if a base URL is not specified, it takes the request URL as an absolute URL for the current host:port combination. In my specific case, when I do a request to `/product/abca`, I expect the request to go to `http://localhost:3000/product/abca`.

### Actual behaviour
Currently it seems to be a weird hybrid as since I didn't specify a base URL and my requests are relative, it appends the port and probably has an empty hostname. Basically currently we have a strange behavior, which possibly should throw an error if we don't want to support empty base URL + relative requests.

### Steps to reproduce
1. let restClient: rc.RestClient = new rc.RestClient('frontend');
2.  let response: rc.IRestResponse<Array<Product>> = await restClient.get<Array<Product>>('/product/' + searchString);

### Logs
Not relevant.

etiago:
I really like the idea behind `typed-rest-client` but as it stands it makes it unusable for a backend+frontend combo unless I hardcode the host where my application is running (which can vary per deployment). Ideally I would like an HTTP client which will just take the current host and pick it up from there when using relative paths.

stephenmichaelf:
Thanks for reaching out about this!

I want to make sure I understand: you are saying when you have the library running on a machine you want it to use the url of the machine it's running on as the base if you don't pass one in? So it can call in to itself? Can't you dynamically get the URL at runtime and set it before you make your request? Or if you want you could do a PR with the change :)

etiago:
Now I realize I didn't explain the problem I'm getting as well as I thought I did...!

[...]

PR #67, opened 2018-04-02
Removes the setting of the default port if one isn't specified.

etiago:
This fixes behavior where a baseUrl isn't defined and we expect the
client to fetch based on the current root url whilst running in the
browser.

Referenced in issue #66 .


etiago:
Unfortunately when I visit https://cla.opensource.microsoft.com/ I don't see any way for signing the CLA digitally... how can I sign it in order to get this merged?

stephenmichaelf:
It looks like the signing requests were met based on [this PR](https://github.com/Microsoft/typed-rest-client/pull/69).

stephenmichaelf:
I am a bit confused, you're saying this is just in the case where there is no base url and it's a relative path? Could you add some unit tests for all scenarios (base/relative and port/no port)? We have to be careful with compat as current users of the lib could be relying on the default port.

Issue #68, opened 2018-04-08
Dates in return types are deserialized as strings 


etiago:
### Environment
Node version: 8.10
Npm version: 5.8.0
OS and version: MacOS 10.13.3
typed-rest-client version: stable

### Issue Description
When using a type for the response which has Date as type for a field, that field ends up being a string, thus violating the type.

### Expected behaviour
If I have a field which is a Date, I expect the deserialization to fail whenever the returned type isn't a date, or if the type does represent a parseable Date, that the field gets correctly transformed into a Date.

### Actual behaviour
Instead of a Date, we get string. This means that any code that I write expecting there to be a Date, becomes type-unsafe as all the Date methods are not available in string objects.

### Steps to reproduce
1. Create type class with Date field.
2. Make request to endpoint which returns a string-encoded standard JSON Date for that field.
3. Verify that none of the Date methods are available in the deserialized object.

As I understand, this may be tricky to solve as `JSON.parse` will see a string and leave it as a string. However, we could possibly introspect the type of the response and handle the Date manually.

bryanmacfarlane:
@damccorm - good issue.  think through compat on potential fixes.

PR #69, opened 2018-04-08
Tries to revive Date types from the JSON

etiago:
I didn't write tests yet as I wanted to brainstorm about this with the developers first. This relates to issue #68 and tries to fix it by using a reviver when parsing the JSON to try and revive any string as a Date. Unfortunately there doesn't seem to be a better option in Javascript.

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -172,6 +172,20 @@ export class RestClient {
         return headers;
     }
 
+    private static dateTimeReviver(key: any, value: any): any 
+    {
+        if (typeof value === 'string')
+        {
+            let a = new Date(value);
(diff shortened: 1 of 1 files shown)


damccorm:
While I see why this might be useful, I'm concerned about a couple things. First, and most importantly, it seems like this could cause serious issues with backwards compatibility - if people are expecting a string and that suddenly shifts to a date that would cause problems. Second, albeit less importantly, this could cause issues with strings that look like dates but aren't meant to be dates.

Overall, it seems to me that this has the potential to cause more issues than its worth, especially since you should be able to just manually do the conversions on your end without too much trouble.

PR #73, opened 2018-07-09
Pass headers to client when passed through constructor

### LINKED ISSUE #65: Headers on RequestOptions in constructor HttpClient not supported
### Environment
typed-rest-client version: 1.0.7

### Issue Description
I can instantiate the HttpClient with default headers as follows: 

this.httpClient = new HttpClient("myUserAgent", [basicHandler], {headers: {"Accept": "application/json"}});

But it doesn't send these headers to the endpoint. In fact, it doesn't seem to use them at all. 

When I send the Accept header as additionalHeader on the get method, everything works as expected
response = await this.httpClient.get("https://example.com/clients", { "Accept": "application/json"});

### Expected behaviour
The default headers, defined in the constructor, should be merged with the additional headers sent along with the get/post/... method.

### Actual behaviour
No default headers are sent to the endpoint.

stephenmichaelf:
This change fixes a bug where headers passed to the HttpClient constructor don't get passed through.

When passing headers in both the constructor and the method, the method parameter will win.

Resolves #65.

Note that headers passed to the request method is now nullable. Since this is more flexible than the current setup I bumped just the minor.

### DIFF
--- a/test/httptests.ts
+++ b/test/httptests.ts
@@ -154,7 +154,6 @@ describe('Http Tests', function () {
     });
 
     it('sends headers with http requests', async() => {
-        this.timeout(5000);
         // calls to /headers returns something like the following (in the body)
         // {"headers":{"Accept":"application/json","Connection":"close","Host":"httpbin.org","User-Agent":"typed-rest-client-tests"}}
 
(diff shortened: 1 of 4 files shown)


test/httptests.ts
```diff
@@ -14,7 +14,7 @@ describe('Http Tests', function () {
     let _httpbin: httpm.HttpClient;
 
     before(() => {
-        _http = new httpm.HttpClient('typed-test-client-tests');
+        _http = new httpm.HttpClient('typed-rest-client-tests'/*, null, {proxy: { proxyUrl: "http://127.0.0.1:8888" } }*/);
```
bryanmacfarlane: commented out code?

stephenmichaelf: Yeah it's just for when we want to use Fiddler to trace things. Left it commented so we can uncomment just for that testing case. Happy to remove if it's confusing.

bryanmacfarlane: the other option is to formalize it - make the tests respect proxy envvar

stephenmichaelf: Do we have a normal process that we follow for setting them? I can see respecting the env var for the lib itself but for tests it makes sense to me to be explicit/easily configurable. Otherwise they could pass on my machine and fail on yours if the env var is set.

stephenmichaelf: Adding offline discussion notes: we will add a custom environment variable to manage this.

PR #75, opened 2018-08-23
Handle 303 SeeOther redirect.

williambernardet:
Implementation is missing 303 redirect support.

I am facing issue downloading a file from a resource which is using this redirect to point to the actual resource.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -13,6 +13,7 @@ export enum HttpCodes {
     MultipleChoices = 300,
     MovedPermanently = 301,
     ResourceMoved = 302,
+    SeeOther = 303,
     NotModified = 304,
     UseProxy = 305,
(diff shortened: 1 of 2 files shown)


bryanmacfarlane:
This looks like a good addition.  Thanks so much.  @stephenmichaelf - agree?

bryanmacfarlane:
@williambernardet - cla ok?

williambernardet:
Just signed it

stephenmichaelf:
This looks great! Thanks for taking the time to make a contribution @williambernardet :)

stephenmichaelf:
Could we add a test where it redirects to a page that gets a 404 or something? To make sure failure cases are handled correctly after redirect.

bryanmacfarlane:
released as 1.0.9

PR #77, opened 2018-09-04
Updated tests to use Nock instead of httpBin

### LINKED ISSUE #72: Replace httpbin with nock
Investigate replacing httpbin requests with Nock.

https://github.com/nock/nock

Would give us more control. Easier to test authentication too?

damccorm:
Switched tests to use nock to spoof Http requests instead of connecting to httpBin. Eliminates timeout failures for tests and allows offline testing. Added tests to ensure that response bodies are included when error is thrown.

Fixes #72 and closes Microsoft/vsts-node-api#34

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,25 @@ latest preview:
 npm install typed-rest-client@preview --save
 ```
 
+# Build
+
+Build:  
+```bash
+$ npm run build
+```
+
+# Test
+
+To run all tests:
+```bash
+$ npm test
+```
+
+To just run unit tests:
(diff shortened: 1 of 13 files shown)


stephenmichaelf:
Looks good. Can we add tests for NTLM and any other missing auth handlers? Also same question here as in the other repo for @bryanmacfarlane about L0 and L2 tests.

stephenmichaelf:
Small thing about Typescript: if you create a variable and it isn't going to change you can use "const" instead of let.

PR #80, opened 2018-09-11
Initial pass at linting

### LINKED ISSUE #39: Linting and strict compilation

damccorm:
Sets up the ability to do more thorough incremental linting, I turned off anything that was going to be a more significant change (though I did get some low hanging fruit). From the options that are currently turned off, we can pick which ones are important and/or easy and go after those ones.

Fixes #39


stephenmichaelf:
This is awesome, thanks!

The biggest thing we need to spend time thinking through is what linting rules we want to have. I think this change will set the precedent for our other open source repos (the ones we own) so it's worthwhile to review the rules and make sure we are all on the same page. I generally like to err on the side of being more strict. We pay the cost once and then it's only per added change in the future.

@bryanmacfarlane

damccorm:
I went ahead and took a couple more smaller passes, I feel pretty good about the set of rules I ended up enabling. Thoughts? If we're happy with this, I could bring the same rules to some of the other projects.

tslint.json
```diff
@@ -0,0 +1,283 @@
+{
```
nschonni: Would it be "better" to `extend` the base set and then disable the ones that don't apply https://github.com/Microsoft/tslint-microsoft-contrib#which-rules-should-i-turn-on

damccorm: Yes, that seems much cleaner, thanks for the recommendation. I've updated accordingly.

stephenmichaelf:
@damccorm Is there an easy way to for us to get a list of rules that we enforce for us to look over?

damccorm:
There's not a fantastic way to do this that I know of, the enforced rules are basically the ones [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/recommended_ruleset.js) except for the ones listed in the tslint file (there's about 25). Alternately, you can look at the ts lint file in [this commit](https://github.com/Microsoft/typed-rest-client/pull/80/commits/aefbc978de2c67c9c8737c226455760ac04a3f79), which comes pretty close to explicitly naming all of them, though a couple of the ones in that file that were listed as true are probably no longer enforced. 

The format in use just ended up being better in terms of rooting out useless rules and helping focus in on only recommended ones, that did come with the tradeoff of being able to easily determine which rules are in use (it does a good job of showing which ones we're intentionally ignoring though).

mmajcica:
Any updates on this? It will be beneficial also for contributors who like me tend to press  SHIFT + ALT+ F often. :)

damccorm:
Closing here and moving to #128

PR #83, opened 2018-09-13
Adding optional date reviver

### LINKED ISSUE #68: Dates in return types are deserialized as strings

damccorm:
Builds off of @etiago's closed PR https://github.com/Microsoft/typed-rest-client/pull/69, makes it optional instead of mandatory to fix issues with back compat, and adds tests.

Fixes #68

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -21,7 +21,7 @@ export interface IRequestOptions {
 
     responseProcessor?: Function,
     //Dates aren't automatically deserialized by JSON, this adds a date reviver to ensure they aren't just left as strings
-    reviveDates?: boolean
+    deserializeDates?: boolean
 }
 
 export class RestClient {
@@ -205,7 +205,7 @@ export class RestClient {
             try {
                 let contents: string = await res.readBody();
                 if (contents && contents.length > 0) {
-                    if (options && options.reviveDates) {
+                    if (options && options.deserializeDates) {
                         obj = JSON.parse(contents, RestClient.dateTimeReviver);
                     } else {
                         obj = JSON.parse(contents);
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -19,7 +19,9 @@ export interface IRequestOptions {
     // since accept is defaulted, set additional headers if needed
     additionalHeaders?: ifm.IHeaders,
 
-    responseProcessor?: Function
+    responseProcessor?: Function,
+    //Dates aren't automatically deserialized by JSON, this adds a date reviver to ensure they aren't just left as strings
+    reviveDates?: boolean
 }
 
 export class RestClient {
@@ -172,6 +174,17 @@ export class RestClient {
         return headers;
     }
 
+    private static dateTimeReviver(key: any, value: any): any {
+        if (typeof value === 'string'){
+            let a = new Date(value);
+            if (!isNaN(a.valueOf())) {
+                return a;
+            }
+        }
+
+        return value;
+    }
+
     private async _processResponse<T>(res: httpm.HttpClientResponse, options: IRequestOptions): Promise<IRestResponse<T>> {
         return new Promise<IRestResponse<T>>(async (resolve, reject) => {
             const statusCode: number = res.message.statusCode;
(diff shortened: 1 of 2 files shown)


etiago:
Thanks for this! I closed mine as I'm no longer using the library and wasn't going to have the time to follow up, but it's looking good 🙏

test/units/resttests.ts
```diff
@@ -64,6 +64,28 @@ describe('Rest Tests', function () {
...
 
+    it('gets a resource and correctly revives its Date property', async() => {
+        const dateObject: Date = new Date();
+        nock('http://microsoft.com')
+            .get('/date')
+            .reply(200, {
+                json: {dateProperty: dateObject.toDateString()}
+            });
+        const restRes: restm.IRestResponse<HttpData> = await _rest.get<HttpData>('http://microsoft.com/date', {reviveDates: true});
+        assert(restRes.result && restRes.result.json.dateProperty.getTime != undefined);
```
stephenmichaelf: assert.equals. Also, can we be more specific with the test? let's use a hard coded datetime and check all the properties of it.

test/units/resttests.ts
```diff
@@ -64,6 +64,28 @@ describe('Rest Tests', function () {
...
+
+    it('gets a resource and doesn\'t revive its non-Date property', async() => {
+        const nonDateObject: string = "stringObject";
+        nock('http://microsoft.com')
+            .get('/date')
+            .reply(200, {
+                json: {nonDateProperty: nonDateObject}
+            });
+        const restRes: restm.IRestResponse<HttpData> = await _rest.get<HttpData>('http://microsoft.com/date', {reviveDates: true});
+        assert(restRes.result && typeof(restRes.result.json.nonDateProperty) == 'string' && restRes.result.json.nonDateProperty == "stringObject");
```
stephenmichaelf: assert.equals. Check response value too.

damccorm: Fixed!

test/units/resttests.ts
damccorm: Fixed!

lib/RestClient.ts
```diff
@@ -19,7 +19,9 @@ export interface IRequestOptions {
     // since accept is defaulted, set additional headers if needed
     additionalHeaders?: ifm.IHeaders,
 
-    responseProcessor?: Function
+    responseProcessor?: Function,
+    //Dates aren't automatically deserialized by JSON, this adds a date reviver to ensure they aren't just left as strings
+    reviveDates?: boolean
```
stephenmichaelf: Is revive a common name for this process? Deserialize might be more helpful to consumers to know what it's doing?

stephenmichaelf: Or maybe there's something else. Revive is a bit weird, may be just me.

damccorm: Most of the documentation I've seen refers to it as a reviver (e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse). Not sure why, maybe people were encoding/decoding dates in different formats and they wanted to make clear that this wasn't a standard part of their deserialization?

Nonetheless, I agree deserialize is more intuitive to me and maybe more accurate in this case since we aren't letting them provide their own custom date encoding/decoding. I'll switch it over.

PR #84, opened 2018-09-13
Support IE11 by targeting es5

### LINKED ISSUE #62: Support IE11 - target es5
Hi again,
this is mainly a question/request.

### Issue Description
We currently use the library for browser usage and as we would like to be able and support IE11 the current npm package is not compatible as the compilation is targeted for es2015.
(The issue is that after compilation the files keep the class references which are not supported by IE11).

### Solution
We have tried to work a solution with that and we have forked and updated the required files to target es5.

Our script works awesome with this change.

You can view the fork and branch here

https://github.com/baxevanis/typed-rest-client/tree/feature/support_es5

Please be aware that we also install the package through npm and github and that is why we had to add some additional configuration in the package.json (prepare and postinstall) to ensure npm can properly install the package.

### Questions
So I have two questions:

1. If we are to make a pull request with the target change to es5, would you consider accepting this change or your script relies on using es2015 explicitly?

2. Can we publish a new npm package ex: typed-rest-client-es5 that will support es5? (by doing this we can clean up the additional changes that are required to install the package from github).

Thank you
Nikolaos

damccorm:
I think this should allow us to target es5. The only things we use that can't be automatically converted by typescript are Promise and Object.assign, both of which I polyfilled.

Fixes #62

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -4,9 +4,12 @@
 import url = require("url");
 import http = require("http");
 import https = require("https");
+import objectAssign = require('object-assign');
 import tunnel = require("tunnel");
 import ifm = require('./Interfaces');
 import fs = require('fs');
+
+//require('es6-promise').polyfill();
 
 export enum HttpCodes {
     OK = 200,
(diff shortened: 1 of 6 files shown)


bryanmacfarlane:
lets talk through before we commit

erwinsmit:
@bryanmacfarlane 
Any update on this?

stephenmichaelf:
@erwinsmit We have not had a chance to discuss this change yet, we will update here when we do! Will try to discuss it in the next day or so.

damccorm:
We're still discussing this on our end, will try to come to a decision soon. If this is blocking you, I think Babel can help in the meantime - https://babeljs.io/docs/en/

damccorm:
Ok, after talking this over a little more, I think we're going to hold off on this for now. We may move forward with this in the future, but if we do we want to put forth a more focused effort with more testing/sample support. If this is an issue for you right now, we recommend forking this repo and copying over this pull request.

Issue #74, opened 2018-08-09
"TypeError: Cannot read property 'resolve' of undefined" when using React


icalvo:
The simplest example failed this error. After debugging I found that `lib/Utils.js` has the failing call: `path.posix.resolve(base.pathname, resultantUrl.pathname)`; but `path.posix` is undefined with my setup. This setup is basically a React web app (meaning that the code is executed on the browser VM rather than by Node), on a Windows 10 box, if that makes a difference. I've simply replaced the call with `path.resolve(base.pathname, resultantUrl.pathname)`, and it works, but I'm not sure if this is altering the outcome for some case.

Stack trace:

    TypeError: Cannot read property 'resolve' of undefined
    at Object.getUrl (http://localhost:3000/static/js/bundle.js:66588:40)
    at RestClient.<anonymous> (http://localhost:3000/static/js/bundle.js:66412:28
    at Generator.next (<anonymous>)    at http://localhost:3000/static/js/bundle.js:66370:71
    at new Promise (<anonymous>)
    at ./node_modules/typed-rest-client/RestClient.js.__awaiter (http://localhost:3000/static/js/bundle.js:66366:12)
    at RestClient.get (http://localhost:3000/static/js/bundle.js:66411:16)
    at TodoStore../src/model/TodoStore.ts.TodoStore.loadTodos (http://localhost:3000/static/js/bundle.js:69754:29)
    at Object../src/App.tsx (http://localhost:3000/static/js/bundle.js:69321:11)
    at __webpack_require__ (http://localhost:3000/static/js/bundle.js:679:30)
    at fn (http://localhost:3000/static/js/bundle.js:89:20)
    at Object../src/index.tsx (http://localhost:3000/static/js/bundle.js:69614:63)
    at __webpack_require__ (http://localhost:3000/static/js/bundle.js:679:30)
    at fn (http://localhost:3000/static/js/bundle.js:89:20)
    at Object.0 (http://localhost:3000/static/js/bundle.js:70021:18)
    at __webpack_require__ (http://localhost:3000/static/js/bundle.js:679:30)
    at ./node_modules/ansi-regex/index.js.module.exports (http://localhost:3000/static/js/bundle.js:725:37)
    at http://localhost:3000/static/js/bundle.js:728:10

stephenmichaelf:
Thanks I will take a look! We haven't fully validated using the lib in the browser.

bryanmacfarlane:
[...]

PR #85, opened 2018-09-14
Added React sample

damccorm:
Simple example designed to show integration of react and library.

Addresses #74


damccorm:
Closing and moving to #86 for compactness/simplicity

Issue #89, opened 2018-10-04
What use var in es2015?


Jungwoo-An:
I think should use const instead of var.

[ntlm.ts](https://github.com/Microsoft/typed-rest-client/blob/master/lib/handlers/ntlm.ts#L8)

damccorm:
Yeah, both _ and ntlm should probably be const. It won't break anything either way, but its good practice to have them as consts. Do you want to submit a pull request for this?

Jungwoo-An:
Yes. I do

PR #91, opened 2018-10-05
Use const instead of var

Jungwoo-An:
Addresses #89


damccorm:
This looks good, thanks for the contribution!

Issue #81, opened 2018-09-12
REST client needs to expose HEAD


stephenmichaelf:
The HTTP client does, we need the REST one to also. It will be consumed in the node api.

stephenmichaelf:
We are going to handle this in generation for the node api.

damccorm:
@stephenmichaelf Do you think this is worth having for other users of the library? If not I'll close my PR, but its a pretty straightforward change I think.

damccorm:
Ok, that makes sense, I'll close the PR.

PR #95, opened 2018-10-16
Expose processResponse method

damccorm:
I don't know if we actually want to do this, but if we're not going to expose HEAD for the restClient, it could be useful to at least expose processResponse for users who want to have a consistent way of accessing responses. A good example of this is azure-devops-node-api, which would probably have to replicate this code if its not exposed in order to solve [this issue](https://github.com/Microsoft/azure-devops-node-api/issues/247).

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -60,7 +60,7 @@ export class RestClient {
         let url: string = util.getUrl(requestUrl, this._baseUrl);
         let res: httpm.HttpClientResponse = await this.client.options(url,
             this._headersFromOptions(options));
-        return this._processResponse<T>(res, options);
+        return this.processHttpResponse<T>(res, options);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Let's discuss. Was there a reason for not exposing HEAD?

damccorm:
Yeah, there was some discussion [here](https://github.com/Microsoft/typed-rest-client/issues/81) and some offline discussion as well (I think?). Basically, we decided HEAD isn't a REST concept and didn't want to expose it since the typed-rest-client is supposed to be a general purpose client, not just servicing the node-api.

Agreed that I'm not sure if we want to do this though. The question I think we have to answer is are we comfortable bumping the major version on any breaking change to processHttpResponse? I think as I consider it now I tend to lean towards no, but obviously that's not where I was 4 months ago. Overall I'd be more inclined to just close this and move on though.

stephenmichaelf:
Let's go with that!

PR #97, opened 2018-10-29
Headers checking in headersFromOptions is now case sensitive

### LINKED ISSUE #96: RestClient additionalHeaders check
In the rest client there is a check if additional headers are specified.

https://github.com/Microsoft/typed-rest-client/blob/7186708f6f5ed96c6c5ca204405915595d5b2149/lib/RestClient.ts#L171

In case user specifies a custom Content-Type header (necessary at example in case we are using uploadStream to upload a file via mutipart/formdata standard [(RFC 2388](https://www.ietf.org/rfc/rfc2388.txt)) the default one will not be added.
The problem is that the check over Content-Type header is case sensitive, so if you added a content-type header, this will not be detected and used correctly.

I would suggest to use a non case sensitive check for this.
If you guys do agree with me, I have a code ready and I'll request a PR.

Thanks

mmajcica:
In case user specifies a custom Content-Type header (necessary at example in case we are using uploadStream to upload a file via mutipart/formdata standard (RFC 2388) the default one will not be added.
The problem is that the check over Content-Type header is case sensitive, so if you added a content-type header, this will not be detected and used correctly.

This makes the comparison case sensitive.

Closes #96

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -168,7 +168,17 @@ export class RestClient {
         headers["Accept"] = options.acceptHeader || "application/json";
 
         if (contentType) {
-            headers["Content-Type"] = headers["Content-Type"] || 'application/json; charset=utf-8';
+            let found: boolean = false;
+
+            for (let header in headers) {
+                if (header.toLowerCase() == "content-type") {
+                    found = true;
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


damccorm:
Looks good, thanks for the contribution!

Issue #98, opened 2018-10-31
Add nock to devDependencies


wesemann:
I think it would be better to add a mocking library which nock is to the devDependencies, because this should not be deployed on production systems. Other opinions about this?

PR #99, opened 2018-10-31
moving nock from dependency to devDependencies

christlang:
#98
* is used only in tests
* will brings a lot transitive dependencies into production environment


damccorm:
Good catch, thanks for the contribution!

PR #101, opened 2018-11-06
Fix getUrl for systems that don't support path.posix

annich-MS:
Some systems (specifically webpack) don't support path.posix. Because using the posix version is only necessary for Node.js on Windows, it seems prudent to fallback to regular path.

Tested with:
Node.js v8.12 on Win 10, OSX, Ubuntu 18.04
Chrome v69 on Win 10, OSX, Ubuntu 18.04
Firefox 62.0.2 on Win 10, Ubuntu 18.04


.gitignore
```diff
@@ -4,6 +4,6 @@ _build
 lib/*.js
 samples/*.js
 samples/*.txt
-test/*.js
```
annich-MS: Looks like gitignore didn't get updated after rearranging the tests folder. Updated for convenience

damccorm: Can we just get rid of testoutput.txt?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.formatOnSave": false
```
annich-MS: Left this in to prevent future VSCode users from accidentally autoformatting (I certainly tripped on it.) You're free to remove it from the pr if you want.

test/tests/httptests.ts
```diff
@@ -118,9 +117,9 @@ describe('Http Tests', function () {
 
     it('pipes a get request', () => {
         return new Promise<string>(async (resolve, reject) => {
-            let file: NodeJS.WritableStream = fs.createWriteStream(sampleFilePath);
-            (await _http.get('https://httpbin.org/get')).message.pipe(file).on('close', () => {
-                let body: string = fs.readFileSync(sampleFilePath).toString();
+            let file = new streamBuffers.WritableStreamBuffer();
```
damccorm: Nit - add type

damccorm: Also, is there a reason to change this? Seems out of scope of the current PR and I'm not really sure what the purpose is

.vscode/settings.json
damccorm: It seems fine to leave it, makes sense.

PR #104, opened 2018-11-08
Fix CI issue

### LINKED ISSUE #103: CI Broken
CI failed on the last commit (before my stream of ugly commits trying to diagnose). Looks like it fails on previous commits as well, so I think this is actually an issue with the CI, not the commit.

damccorm:
Right now, npm test is failing in CI for lower versions of node. This is happening because it is trying to build before testing. This fixes this.

Fixes #103


make.js
```diff
@@ -86,6 +86,20 @@ target.test = function() {
     run('mocha test/tests');
 }
 
+target.testNoBuild = function() {
```
stephenmichaelf: Can we just make test be no build instead of having two?

make.js
```diff
@@ -86,6 +86,20 @@ target.test = function() {
...
+    pushd('test')
+    run('npm install ../_build');
+    popd();
+
+    console.log("-------Unit Tests-------");
+    run('tsc -p ./test/units');
+    run('mocha test/units');
+
+    console.log("-------Other Tests-------");
```
stephenmichaelf: What are "other" tests?

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "node make.js build",
     "test": "node make.js test",
+    "testNoBuild": "node make.js testNoBuild",
```
stephenmichaelf: Does the yaml script need to be updated?

make.js
damccorm: I could go either way on this. I think its nice to build before testing in general because it removes that step. I updated for better modularity though so its less of a big deal. If you feel strongly I'm also ok getting rid of this

make.js
damccorm: Non-nock tests. I don't really know what the best name for them is, the only difference here is that they actually make http calls. They're not really integration tests since they're still only testing individual components, but they also are a little more involved/flaky than the units.

package.json
damccorm: I think I did this above

make.js
stephenmichaelf: Integration is close enough I think.

make.js
stephenmichaelf: Yeah let's get rid of it and keep them separate.

make.js
damccorm: Ok, should be good to go

stephenmichaelf:
Is this still broken? Can we merge?

damccorm:
I think yes to both, I'll merge.

PR #105, opened 2018-12-05
Add headers to rest response

### LINKED ISSUE #87: Propogate headers in RestClient's IRestResponse
### Environment
Node version: 
Npm version: 8.11.4
OS and version: Windows 10
typed-rest-client version: 1.0.9

### Issue Description
Currently, in RestClient, calls to the public functions all receive a httpm.HttpClientResponse object which contains headers but return a promise of an IRestResponse object which does not contain headers.

### Expected behaviour
IRestResponse should contain headers

### Actual behaviour
IRestResponse only contains information pulled from the result and the status code.

### Steps to reproduce

```
let restc: RestClient = new RestClient('rest-samples', 'httpbin.org/');
let restRes: IRestResponse<Object> = await restc.get<Object>('get');
```
Try to get headers from restRes

damccorm:
Fixes #87 by appending headers to what we're returning. While fixing this, ran into an issue with the samples that I fixed.


lib/RestClient.ts
```diff
@@ -9,7 +9,8 @@ import util = require("./Util");
 
 export interface IRestResponse<T> {
     statusCode: number,
-    result: T | null
+    result: T | null,
+    headers: Object
```
stephenmichaelf: I wonder if we should make this of type [IHeaders](https://github.com/Microsoft/typed-rest-client/blob/master/lib/Interfaces.ts#L7)?

lib/RestClient.ts
```diff
@@ -227,6 +229,7 @@ export class RestClient {
                         response.result = obj;
                     }
                 }
+                response.headers = res.message.headers;
```
stephenmichaelf: if (res.message.headers) ?

lib/RestClient.ts
stephenmichaelf: Genuine question... I am not sure :)

damccorm: I originally went with object because that's what it is [here](https://nodejs.org/api/http.html#http_message_headers). Additionally, I don't really see a difference between using IHeaders and Object - isn't a JavaScript object just a collection of string keys mapping to values (aka an IHeader)? I don't think any additional info is conveyed there, all we're really doing by calling it a IHeader is making the user look one more place for a type definition. If anything I'd be more partial to just doing away with the IHeaders type (though we probably shouldn't actually do that because of back-compat).

lib/RestClient.ts
damccorm: I think this is [guaranteed to have the headers property](https://nodejs.org/api/http.html#http_class_http_incomingmessage) so I think we can just expect it (like we do with status code above).

damccorm:
@stephenmichaelf can we merge?

ThomRoux:
According to Rest specs, creating a resource in a Rest API doesn't return any body, but returns a Location header, pointing to the created resource. 
It is imperative that headers are returned in IRestReponse.

Can this be merged?

stephenmichaelf:
Yes let's merge.

PR #124, opened 2019-01-22
Accept error message in body as a string

mmajcica:
Solves #123

PR #127, opened 2019-02-05
Added a test for the statusCode > 299 and text body

mmajcica:
As requested in https://github.com/Microsoft/typed-rest-client/pull/124 I added an integration test that will test the code.


damccorm:
This looks good to me, thanks for following through on this.

PR #131, opened 2019-02-11
Add option for presigned urls

### LINKED ISSUE #126: CredentialHandlers add authorization headers to presigned urls
### Environment
Node version: v8.12.0
Npm version: 6.4.1
OS and version: Windows 10 Build 17763
typed-rest-client version: 

### Issue Description
We have a scenario where we call the artifacts packaging api to download a package file. This api call redirects you to a presigned blobstore url which has the file contents. The problem I am facing is that the headers from the original api call (including the authorization header) are passed along to the redirected URL, which is presigned in this case, and causes a 401. If I remove the auth header in the redirect logic, it works correctly

### Expected behaviour
CredentialHandlers don't add authorization headers to presigned urls.

### Actual behaviour
CredentialHandlers add authorization headers to presigned urls.

### Steps to reproduce
1. Go to any url that redirects to a presigned url.
2. Enable redirection in the client
3. Response is 401

### Possible fixes
1. The prepareRequest method checks the url path to see if it is preauthorized and disables adding authorization handler.
2. An option is provided for the user to set what urls don't need authentication, using this option, the authhandler logic is completely skipped.

### See also
I checked in these files, there is a TODO to fix https://github.com/Microsoft/typed-rest-client/blob/e8cb7c8490a34073e9744afc995df1a088583492/lib/handlers/basiccreds.ts#L16, perhaps its related to this bug

damccorm:
Allow users to whitelist presigned url patterns

Fixes #126

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -341,13 +341,26 @@ export class HttpClient implements ifm.IHttpClient {
         info.options.agent = this._getAgent(requestUrl);
 
         // gives handlers an opportunity to participate
-        if (this.handlers) {
+        if (this.handlers && !this._isPresigned(requestUrl)) {
             this.handlers.forEach((handler) => {
                 handler.prepareRequest(info.options);
             });
         }
 
         return info;
+    }
+
+    private _isPresigned(requestUrl): boolean {
+        if (this.requestOptions && this.requestOptions.presignedUrlPatterns) {
+            const patterns: RegExp[] = this.requestOptions.presignedUrlPatterns;
+            for (let i: number = 0; i < patterns.length; i += 1) {
+                if (requestUrl.match(patterns[i])) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
     }
 
     private _mergeHeaders(headers: ifm.IHeaders) : ifm.IHeaders {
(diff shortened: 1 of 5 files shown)


lib/HttpClient.ts
```diff
@@ -350,6 +350,19 @@ export class HttpClient implements ifm.IHttpClient {
         return info;
     }
 
+    private _isPresigned(requestUrl): boolean {
```
stephenmichaelf: Add type to parameter.

lib/HttpClient.ts
```diff
@@ -350,6 +350,19 @@ export class HttpClient implements ifm.IHttpClient {
...
+        if (this.requestOptions && this.requestOptions.presignedUrlPatterns) {
+            const patterns: RegExp[] = this.requestOptions.presignedUrlPatterns;
+            for (let i: number = 0; i < patterns.length; i += 1) {
```
stephenmichaelf: i++?

stephenmichaelf:
@aasim Will this solve your needs? Will we always know the structure of the URLs to skip?

lib/HttpClient.ts
damccorm: Yeah, I did this because the ESLint style guide prefers +=, but I think that's actually just outside the context of loops. I'll go with ++

damccorm:
@aasim will this solve your use case in #126? @stephenmichaelf I'd probably want to merge this regardless, it seems like a useful thing to support.

PR #134, opened 2019-03-06
Enable retries on some 500 codes on reads

damccorm:
This should enable https://github.com/Microsoft/azure-devops-node-api/issues/288 to be done more easily


lib/HttpClient.ts
```diff
@@ -186,48 +198,62 @@ export class HttpClient implements ifm.IHttpClient {
         }
 
         let info: RequestInfo = this._prepareRequest(verb, requestUrl, headers);
-        let response: HttpClientResponse = await this.requestRaw(info, data);
 
-        // Check if it's an authentication challenge
-        if (response && response.message && response.message.statusCode === HttpCodes.Unauthorized) {
-            let authenticationHandler: ifm.IRequestHandler;
+        // Only perform retries on reads since writes may not be idempotent.
+        let numTries = (this._allowRetries && HttpWriteOptions.indexOf(verb) == -1) ? this._maxRetries + 1 : 1;
```
stephenmichaelf: numTriesRemaining?

stephenmichaelf: numTriesRemaining?

lib/HttpClient.ts
```diff
@@ -186,48 +198,62 @@ export class HttpClient implements ifm.IHttpClient {
...
-                break;
+                // let's make the request with the new redirectUrl
+                info = this._prepareRequest(verb, redirectUrl, headers);
+                response = await this.requestRaw(info, data);
+                redirectsRemaining--;
             }
 
-            // we need to finish reading the response before reassigning response
-            // which will leak the open socket.
-            await response.readBody();
```
stephenmichaelf: I think we need to keep this.

lib/HttpClient.ts
```diff
@@ -38,6 +38,8 @@ export enum HttpCodes {
 }
 
 const HttpRedirectCodes: number[] = [HttpCodes.MovedPermanently, HttpCodes.ResourceMoved, HttpCodes.SeeOther, HttpCodes.TemporaryRedirect, HttpCodes.PermanentRedirect];
+const HttpRetryCodes: number[] = [HttpCodes.BadGateway, HttpCodes.ServiceUnavailable, HttpCodes.GatewayTimeout];
+const HttpWriteOptions: string[] = ['DELETE', 'POST', 'PUT', 'PATCH'];
```
stephenmichaelf: Do we need to expose this?

lib/HttpClient.ts
damccorm: Yeah, its still there, the diff is just displaying differently. Its on line 243 and occurs in the same place relative to the original code.

lib/HttpClient.ts
damccorm: I don't think so - what would it be used for? I'd rather expose less if we don't have a clear use case

stephenmichaelf:
I think we have tests for redirect, can we add some for this too?

damccorm:
Good call, added a test which actually caught an issue!

lib/HttpClient.ts
```diff
@@ -186,48 +198,62 @@ export class HttpClient implements ifm.IHttpClient {
...
-            let authenticationHandler: ifm.IRequestHandler;
+        // Only perform retries on reads since writes may not be idempotent.
+        let numTriesRemaining = (this._allowRetries && HttpWriteOptions.indexOf(verb) == -1) ? this._maxRetries + 1 : 1;
 
-            for (let i = 0; i < this.handlers.length; i++) {
-                if (this.handlers[i].canHandleAuthentication(response)) {
-                    authenticationHandler = this.handlers[i];
-                    break;
+        let response: HttpClientResponse;
+        while (numTriesRemaining > 0) {
```
stephenmichaelf: We should add a sleep/exponential backoff.

lib/HttpClient.ts
stephenmichaelf: Delete is probably fine. Can we change this to opt in list instead of opt out?

lib/HttpClient.ts
```diff
@@ -38,6 +38,8 @@ export enum HttpCodes {
 }
 
 const HttpRedirectCodes: number[] = [HttpCodes.MovedPermanently, HttpCodes.ResourceMoved, HttpCodes.SeeOther, HttpCodes.TemporaryRedirect, HttpCodes.PermanentRedirect];
+const HttpRetryCodes: number[] = [HttpCodes.BadGateway, HttpCodes.ServiceUnavailable, HttpCodes.GatewayTimeout];
```
stephenmichaelf: HttpResponseRetryCodes.

lib/HttpClient.ts
stephenmichaelf: Maybe rename to RetryableHttpVerbs?

test/units/httptests.ts
```diff
@@ -422,4 +422,19 @@ describe('Http Tests with keepAlive', function () {
         let res: httpm.HttpClientResponse = await _http.options('http://microsoft.com');
         assert(res.message.statusCode == 200, "status code should be 200");
     });
+
+    it('handles retries correctly', async() => {
+        _http = new httpm.HttpClient('typed-test-client-tests', null, {allowRetries: true, maxRetries: 1});
```
stephenmichaelf: Lets do 4 or 5 so we make sure there are no off by one issues at a value of 1.

lib/HttpClient.ts
```diff
@@ -186,48 +198,62 @@ export class HttpClient implements ifm.IHttpClient {
...
+                    break;
+                }
 
-            const redirectUrl: any = response.message.headers["location"];
-            if (!redirectUrl) {
-                // if there's no location to redirect to, we won't
-                break;
+                // we need to finish reading the response before reassigning response
+                // which will leak the open socket.
+                await response.readBody();
```
stephenmichaelf: I think we need to also do this in the outer retry loop?

stephenmichaelf:
Looks like tests are failing.

damccorm:
From offline discussion, this is a separate known issue with tests timing out and can be ignored.

damccorm:
Accidentally merged this, has since been reverted and moved to #137

Issue #90, opened 2018-10-04
Add nock tests for handlers


damccorm:
Should test all of them, but especially NTLM since it doesn't do preauth and is pretty fragile

hamzahejja:
@damccorm I'm able to pre-plan the unit tests for the Basic, Personal Access Token and Bearer Token Authentication schemes, as they all perform pre-auth by Setting the `Authorization` header on the outgoing request. However, couldn't imagine how to carry-on the test for the NTLM auth, especially as it involves CLIENT-SERVER Handshaking and is quite complex compared to other schemes, So i bumped into a wall while thinking to write the test. Any suggestions? thoughts ?

hamzahejja:
@damccorm I've managed to pre-plan and setup the needed tests for **_Basic, Bearer and Personal Access Token (PAT)_** Authentication schemes, by setting nock to match the `Authorization` header on the outgoing HTTP request. However, I'm wondering how to carry out the test for the NTLM Authentication scheme, as it is more complex, doesn't perform any Pre-auth step on the request and involves Client-Server Handshaking. Any thoughts on this? suggestions?

damccorm:
@hamzahejja the NTLM one is definitely the hardest - I think you should still be able to do it more or less the same way though, right? You would just need to mock multiple sets of HTTP requests with different parameters, which nock _should_ allow.

With that said, feel free to open a PR on the other 3 handlers if you have them - happy to get those merged in and we can try to work on the ntlm one separately.

hamzahejja:
@damccorm I have just had another look into the tests which currently exists,  **PAT and Basic** Authentication schemes have nock tests set for them:
_Basic Auth_: https://github.com/microsoft/typed-rest-client/blob/master/test/units/httptests.ts#L72-L99
_Personal Access Token (PAT)_: https://github.com/microsoft/typed-rest-client/blob/master/test/units/httptests.ts#L130-L160

Which leaves out the Bearer Token Auth. scheme and the NTLM, right ?

damccorm:
Yep, that sounds right!

hamzahejja:
@damccorm
I have set the following nock test for the NTLM Authentication Handler, depending on the example given **_BOTTOM OF PAGE_**, with the following settings:
`username`: **Zaphod**
`password`: **Beeblebrox**
`workstation` (i.e. hostname): **LightCity**
`domain` (i.e. NTDomain): **Ursa-Minor**
```
[...]

PR #135, opened 2019-03-08
Added private token handler

mmajcica:
Several tools, like Gitlab, do support authentication through the Private-Token header field and related PAT.
Check this for further info https://docs.gitlab.com/ee/api/#personal-access-tokens 

I created a new handler that will do just that.

### DIFF
--- a/lib/handlers/privatetoken.ts
+++ b/lib/handlers/privatetoken.ts
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import ifm = require('../Interfaces');
+
+export class PrivateTokenCredentialHandler implements ifm.IRequestHandler {
+    token: string;
+
+    constructor(token: string) {
+        this.token = token;
+    }
+
+    prepareRequest(options:any): void {
+        options.headers['Private-Token'] = this.token;
+    }
+
+    // This handler cannot handle 401
+    canHandleAuthentication(response: ifm.IHttpClientResponse): boolean {
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


damccorm:
I'm not very familiar with this so I need to do a little more research on this to understand its usage so that I can give a good review. In the meantime, a couple things that could be helpful:

1) I'm struggling to find any tools besides GitLab that use the `Private-Token` header. If its just GitLab I'd probably be more inclined to just encourage users to use the additional-headers parameter and pass it in, but it sounds like other things use it as well - could you point me to an example or 2?
2) An export needs to be added [here](https://github.com/Microsoft/typed-rest-client/blob/e2b762f6849d2b2b4eb51198efaf999c411deb06/lib/Handlers.ts).
3) Could you add a simple test to confirm this does what its supposed to? I know we don't have testing for other handlers, but [we want to](https://github.com/Microsoft/typed-rest-client/issues/90) and I'd rather start that now. It probably belongs in a separate file [test/units](https://github.com/Microsoft/typed-rest-client/tree/master/test/units)/handlers

mmajcica:
@damccorm you are tight, not many tools are still using it. Aside of Gitlab, there is gitswarm that I do know about. However, it is not much code and can be an option. If you think it is a clutter, I'm also fine to keep it only in my projects.
I do agree about the tests and I would like to give it a try. Once you decide if to keep this, then I'll add test to the same PR, otherwise I'll go with a separate one for the tests.

damccorm:
I think lets keep this separate from the library. Since we allow people to pass in custom handlers I think it makes sense to keep this lib to more standard/universal handler types so that we can keep it tight and small. With that said, if you disagree/think this will be useful for other people you could also publish it as your own stand-alone npm package, I think that's probably a good pattern for us to use with handlers in general.

mmajcica:
I agree with that.

PR #137, opened 2019-03-08
Enable retries on some 500 codes on reads

damccorm:
This should enable Microsoft/azure-devops-node-api#288 to be done more easily


lib/HttpClient.ts
```diff
@@ -38,6 +38,10 @@ export enum HttpCodes {
 }
 
 const HttpRedirectCodes: number[] = [HttpCodes.MovedPermanently, HttpCodes.ResourceMoved, HttpCodes.SeeOther, HttpCodes.TemporaryRedirect, HttpCodes.PermanentRedirect];
+const HttpResponseRetryCodes: number[] = [HttpCodes.BadGateway, HttpCodes.ServiceUnavailable, HttpCodes.GatewayTimeout];
+const RetryableHttpVerbs: string[] = ['OPTIONS', 'GET', 'DELETE', 'HEAD'];
+const ExponentialBackoffCeiling = 10;
+const ExponentialBackoffSlotTime = 5;
```
stephenmichaelf: Not sure what slot time means?

lib/HttpClient.ts
```diff
@@ -186,48 +200,68 @@ export class HttpClient implements ifm.IHttpClient {
...
 
         let info: RequestInfo = this._prepareRequest(verb, requestUrl, headers);
-        let response: HttpClientResponse = await this.requestRaw(info, data);
 
-        // Check if it's an authentication challenge
-        if (response && response.message && response.message.statusCode === HttpCodes.Unauthorized) {
-            let authenticationHandler: ifm.IRequestHandler;
+        // Only perform retries on reads since writes may not be idempotent.
+        let maxTries = (this._allowRetries && RetryableHttpVerbs.indexOf(verb) != -1) ? this._maxRetries + 1 : 1;
+        let numTries = 0;
```
stephenmichaelf: Add type.

lib/HttpClient.ts
```diff
@@ -186,48 +200,68 @@ export class HttpClient implements ifm.IHttpClient {
...
+                redirectsRemaining--;
             }
 
-            if (authenticationHandler) {
-                return authenticationHandler.handleAuthentication(this, info, data);
-            }  
-            else {
-                // We have received an unauthorized response but have no handlers to handle it.
-                // Let the response return to the caller.
+            if (HttpResponseRetryCodes.indexOf(response.message.statusCode) == -1) {
```
stephenmichaelf: && check RetryableHttpVerbs? Only retry if in there. Let's add tests to make sure we don't retry on not retryable verbs, regardless of response.

lib/HttpClient.ts
```diff
@@ -186,48 +200,68 @@ export class HttpClient implements ifm.IHttpClient {
...
-        while (HttpRedirectCodes.indexOf(response.message.statusCode) != -1
-               && this._allowRedirects
-               && redirectsRemaining > 0) {
+            numTries += 1
 
-            const redirectUrl: any = response.message.headers["location"];
-            if (!redirectUrl) {
-                // if there's no location to redirect to, we won't
-                break;
+            if (numTries != maxTries) {
```
stephenmichaelf: Can we use < instead of != ?

lib/HttpClient.ts
```diff
@@ -501,4 +535,10 @@ export class HttpClient implements ifm.IHttpClient {
 
         return bypass;
     }
+
+    private _performExponentialBackoff(retryNumber: number): Promise<void> {
+        retryNumber = Math.min(ExponentialBackoffCeiling, retryNumber);
+        const ms: number = ExponentialBackoffSlotTime/Math.pow(2, retryNumber);
```
stephenmichaelf: We should set some max cap on this.

damccorm: That's what the `Math.min` expression does. It sets the cap at `SlotTime*(2^Ceiling)`

lib/HttpClient.ts
damccorm: We don't need to check `RetryableHttpVerbs` because of this expression above: `let maxTries = (this._allowRetries && RetryableHttpVerbs.indexOf(verb) != -1) ? this._maxRetries + 1 : 1;`

If its not in `RetryableHttpVerbs`, `maxTries` will equal 1 and it will fall out of the loop.

damccorm: Added more testing

damccorm:
@stephenmichaelf when you get a chance could you take one last look at this? I think it should be good to go.

lib/HttpClient.ts
```diff
@@ -186,48 +200,68 @@ export class HttpClient implements ifm.IHttpClient {
         }
 
         let info: RequestInfo = this._prepareRequest(verb, requestUrl, headers);
-        let response: HttpClientResponse = await this.requestRaw(info, data);
 
-        // Check if it's an authentication challenge
-        if (response && response.message && response.message.statusCode === HttpCodes.Unauthorized) {
-            let authenticationHandler: ifm.IRequestHandler;
+        // Only perform retries on reads since writes may not be idempotent.
+        let maxTries: number = (this._allowRetries && RetryableHttpVerbs.indexOf(verb) != -1) ? this._maxRetries + 1 : 1;
```
stephenmichaelf: I think its more clear to put the retryable http verbs part in the while loop. Otherwise you have to read that then look back up here.

lib/HttpClient.ts
```diff
@@ -186,48 +200,68 @@ export class HttpClient implements ifm.IHttpClient {
...
+                // If not a retry code, return immediately instead of retrying
                 return response;
             }
-        }
 
-        let redirectsRemaining: number = this._maxRedirects;
-        while (HttpRedirectCodes.indexOf(response.message.statusCode) != -1
-               && this._allowRedirects
-               && redirectsRemaining > 0) {
+            numTries += 1
```
stephenmichaelf: Missing semicolon?

lib/HttpClient.ts
damccorm: We need the contents of the loop to run at least once regardless of the verb though. So if we move it into the while condition we'd have to do some weird special casing that would make it more confusing IMO.

lib/HttpClient.ts
damccorm: Good catch

stephenmichaelf:
Going to approve, my main concern was making sure we do readBody when we retry. We had a bug before where that didn't happen for redirect and it left a socket open and caused issues. It seems like we feel better on that through the tests.

PR #140, opened 2019-03-26
add support FormData

josheplibra:


stephenmichaelf:
Thanks for taking the time to make a contribution. Unfortunately it's unlikely we will add this to the code.

As was mentioned in your original question the best way to do this is to keep the setup in client code then serialize it to a string and pass to the HttpClient. It's not something that would be the responsibility of the lib to know about.

@damccorm Will be adding an example.

lib/HttpClient.ts
```diff
@@ -347,11 +348,17 @@ export class HttpClient implements ifm.IHttpClient {
         }
 
         if (data && typeof (data) !== 'string') {
-            data.on('close', function () {
+            if (util.isFormData(data)) {
+                req.write(data);
```
damccorm: Just a follow up to Stephen's comment, but Node's `http.ClientRequest`'s `write` function only takes in a `string | Buffer` (see docs [here](https://nodejs.org/api/http.html#http_request_write_chunk_encoding_callback)). So we should only be taking `string | Buffer` types, not FormData (and I don't think this implementation will work as such).

damccorm: I'll add an example of how to do this, but the basic gist of it is that you need to convert your formData object into either a string or Buffer to be compatible with this library (and Node's `ClientRequest`)

PR #156, opened 2019-06-06
Allow no user agent to be passed

### LINKED ISSUE #154: If userAgent is null or not provided it is still added to the headers
The 'userAgent' param is set in the headers even if it is null or undefined. I would have expected the behavior to be to conditionally add the header only if it is not null.

damccorm:
If they explicitly pass in null or undefined we shouldn't attach the header to avoid issues like #154. In theory this could break someone, but only if all of the following conditions are met:

1) They are currently passing in null or undefined for the userAgent (not just `''`). Note that this is not allowed by our typing, but some JS clients or TS clients without strict typing could in theory be doing this
2) The endpoint they are hitting requires a userAgent header but allows it to be empty

This feels pretty low risk to me so I think a minor version bump should be fine.

Fixes #154

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -5,7 +5,6 @@ import url = require("url");
 import http = require("http");
 import https = require("https");
 import ifm = require('./Interfaces');
-import { parse } from "querystring";
 let fs: any;
 let tunnel: any;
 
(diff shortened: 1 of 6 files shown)


stephenmichaelf:
Can we add tests for this

bryanmacfarlane:
Resolve the conflicts and we can merge

rumblefrog:
Could we also add this to the RestClient?

damccorm:
@rumblefrog sure! Added https://github.com/microsoft/typed-rest-client/pull/193

PR #160, opened 2019-07-09
Consider NO_PROXY environment variable for default proxyBypassHosts List

BobSilent:
Current implementation considers HTTP_PROXY and HTTPS_PROXY in _getProxy as fallback when request does not provide a proxy configuration but system has the variables configured.
unfortunately it ignores the NO_PROXY settings.

With this change the _httpProxyBypassHosts gets initiallized (as fallback) with NO_PROXY environment settings, and gets overwritten by requestOptions.proxy.proxyBypassHost if available.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -78,6 +78,7 @@ export function isHttps(requestUrl: string) {
 enum EnvironmentVariables {
     HTTP_PROXY = "HTTP_PROXY",
     HTTPS_PROXY = "HTTPS_PROXY",
+    NO_PROXY = "NO_PROXY",
 }
 
 export class HttpClient implements ifm.IHttpClient {
@@ -105,6 +106,14 @@ export class HttpClient implements ifm.IHttpClient {
     constructor(userAgent: string, handlers?: ifm.IRequestHandler[], requestOptions?: ifm.IRequestOptions) {
         this.userAgent = userAgent;
         this.handlers = handlers || [];
+        let no_proxy: string = process.env[EnvironmentVariables.NO_PROXY];
+        if (no_proxy) {
+            this._httpProxyBypassHosts = [];
+            no_proxy.split(',').forEach(bypass => {
+                this._httpProxyBypassHosts.push(new RegExp(bypass, 'i'));
+            });
+        }
+
         this.requestOptions = requestOptions;
         if (requestOptions) {
             if (requestOptions.ignoreSslError != null) {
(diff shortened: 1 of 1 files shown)


nimatt:
This would be great. We have an issue with our build server sitting behind a proxy that this would solve.

lib/HttpClient.ts
```diff
@@ -105,6 +106,14 @@ export class HttpClient implements ifm.IHttpClient {
     constructor(userAgent: string, handlers?: ifm.IRequestHandler[], requestOptions?: ifm.IRequestOptions) {
         this.userAgent = userAgent;
         this.handlers = handlers || [];
+        let no_proxy: string = process.env[EnvironmentVariables.NO_PROXY];
+        if (no_proxy) {
+            this._httpProxyBypassHosts = [];
+            no_proxy.split(',').forEach(bypass => {
+                this._httpProxyBypassHosts.push(new RegExp(bypass, 'i'));
```
nimatt: Don't the bypass needs to be processed a bit before turned into a regex?

Can think of a few scenarios

1. If there is a space after the comma it will now be part of the pattern
2. Escape . so that github.com does not match githubFcom
3. Normally adding a domain includes all subdomains e.g. having github.com should include raw.github.com

With this I'm thinking that maybe a version with safe checking of something like
`new RegExp('(^|.+\\.)' + bypass.trim().replace('.', '\\.'))`
Could work?

BobSilent: I was also thinking about this, but i decided to align it with Line 128 
``` javascript
            if (requestOptions.proxy && requestOptions.proxy.proxyBypassHosts) {
                this._httpProxyBypassHosts = [];
                requestOptions.proxy.proxyBypassHosts.forEach(bypass => {
                    this._httpProxyBypassHosts.push(new RegExp(bypass, 'i'));
                });
            }
```

normally in linux environments the no_proxy is not a regex:

> This variable should contain a comma-separated list of domain extensions proxy should not be used for. For instance, if the value of no_proxy is ‘.mit.edu’, proxy will not be used to retrieve documents from MIT.

see http://www.gnu.org/software/wget/manual/html_node/Proxies.html

BobSilent:
> This would be great. We have an issue with our build server sitting behind a proxy that this would solve.

yes, facing a similar problem, i need a proxy for connecting to internet, but no proxy for accessing on prem DevOps Server. 
I already patched some extensions using this package with my provided pull request to be able to use them

jtpetty:
@BobSilent - thank you for your contribution! I am sorry it has taken so long for us to take action on it. It looks good to me.

PR #161, opened 2019-07-25
Adding support for text/plain content type responses

emtencza:
Currently, the RestClient only accepts json/application content type responses, and will return null for text/plain content type responses. This change will set the response result to the raw content if JSON parsing fails (eg contents are not a JSON). This will allow for text/plain content type responses.

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -221,11 +221,6 @@ export class RestClient {
                     } else {
                         obj = JSON.parse(contents);
                     }
-                    // If the response was text/plain, JSON.parse will return undefined
-                    if (obj === undefined) {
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
The typed-rest-client has two layers - rest client which turns a request into a resourceful json object and the http client which is used for raw http requests.  If you're making a request that doesn't return and restful resource object (and instead a file stream, raw string contents etc...) then you should use the http client.

What's the scenario / issue where this is needed and the http client is not suitable?

emtencza:
The http client works well for my scenario. I was trying to get the MIME content of an Outlook message using the rest client. I have switched it to the http client and it works. Thank you and I will abandon this PR.

Issue #167, opened 2019-09-30
Corrupted UTF-8 data in http response


sergeykochanov:
1. Please check our current Issues to see if someone already reported this https://github.com/Microsoft/typed-rest-client/issues
2. Take a look at our Instructions for logging issues https://github.com/Microsoft/typed-rest-client/blob/master/CONTRIBUTING.md#instructions-for-logging-issues

### Environment
Node version: 10.15.0
Npm version: 6.4.1
OS and version: Ubuntu 16.04.5
typed-rest-client version: 0.12.0

### Issue Description
When downloading a http response with long utf-8 content sometimes it changes valid symbol to ��.

### Expected behaviour
Content should be downloaded as is.

### Actual behaviour
One symbol in content is corrupted.

### Steps to reproduce
1. Have http server with large (100kb and more) unicode document
2. Download it using typed-rest-client
3. Check content for �� symbols. 
4. Repeat.

### Logs

```
async function run() {

    let project: string = "xxxx";
   
    while (1) {
        let connection = new vm.WebApi(orgUrl, authHandler);    
        let vstsTask: ta.ITaskAgentApi = await connection.getTaskAgentApi();
        let vstsRelease: ra.IReleaseApi = await connection.getReleaseApi();
        let release: ri.Release = await vstsRelease.getRelease(project, 747);
        if (JSON.stringify(release).includes('��')) {
            console.log("[-] Detected wrong charachter, trying again.")
            break;
        }
        console.log("[+] All right, everyting is fine.")
        await vstsRelease.updateRelease(release, project, 747);
    }  
}
```

bryanmacfarlane:
Re-opening since the fix caused CI to fail (which wasn't working when the PR was created).  Please try another PR.

sergeykochanov:
My bad - I copied JS code from a patch instead of TS. Fixed.

kiikura:
I have the same problem.  
I would like to use the fixed version. When would it be published on npm?

PR #168, opened 2019-09-30
Fix data corruption

sergeykochanov:
See #167


bryanmacfarlane:
Thanks for your contribution!  I assume your repro from the bug doesn't reproduce with the fix, correct?

sergeykochanov:
Yes, i’m using the fix for a few month.

PR #173, opened 2019-10-10
Allowing redirect downgrade from HTTPS to HTTP should be optional

### LINKED ISSUE #165: Forbid Redirect Downgrade from HTTPS to HTTP
## Feature Request

If an HTTPS client request gets downgraded by a redirect to HTTP by some host in the redirect chain, users should have the option to forbid this downgrade in the configuration of the `HttpClient`.

## Impacted Code

https://github.com/microsoft/typed-rest-client/blob/c99dbbeba106d4ca59b62a320923eed9b9a6c223/lib/HttpClient.ts#L239-L256

There is no check to see if the new redirect location is a downgrade from HTTPS to HTTP.

While not strictly a security vulnerability, this does have security implications.

## Use Case

The use case for this is the GitHub Actions toolkit API. See: https://github.com/actions/toolkit/issues/162

The GitHub actions toolkit is used to download other tools using the `downloadTool` API. We'd like to require that users provide SHA-256 checksums for their artifacts if they end up using HTTP instead of HTTPS to prevent MITM attacks against the GH Actions supply chain.

This verification should also be required if somewhere in the redirect chain in downloading the tool, the request is downgraded from HTTPS to HTTP. Currently, the `HttpClient` doesn't support this sort of check.

JLLeitschuh:
Closes #165

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -254,7 +254,7 @@ export class HttpClient implements ifm.IHttpClient {
                 }
                 let parsedRedirectUrl = url.parse(redirectUrl);
                 if (parsedUrl.protocol == 'https:' && parsedUrl.protocol != parsedRedirectUrl.protocol && !this._allowRedirectDowngrade) {
-                    throw new Error("Redirect from HTTPS to HTTP protocol.");
+                    throw new Error("Redirect from HTTPS to HTTP protocol. This downgrade is not allowed for security reasons. If you want to allow this behavior, set the allowRedirectDowngrade option to true.");
                 }
 
                 // we need to finish reading the response before reassigning response
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -91,7 +91,8 @@ export class HttpClient implements ifm.IHttpClient {
     private _httpProxy: ifm.IProxyConfiguration;
     private _httpProxyBypassHosts: RegExp[];
     private _allowRedirects: boolean = true;
-    private _maxRedirects: number = 50
+    private _allowRedirectDowngrade: boolean = false;
+    private _maxRedirects: number = 50;
     private _allowRetries: boolean = false;
     private _maxRetries: number = 1;
     private _agent;
@@ -143,6 +144,10 @@ export class HttpClient implements ifm.IHttpClient {
 
             if (requestOptions.allowRedirects != null) {
                 this._allowRedirects = requestOptions.allowRedirects;
+            }
+
+            if (requestOptions.allowRedirectDowngrade != null) {
+                this._allowRedirectDowngrade = requestOptions.allowRedirectDowngrade;
             }
 
             if (requestOptions.maxRedirects != null) {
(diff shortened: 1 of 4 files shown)


lib/HttpClient.ts
```diff
@@ -241,18 +247,22 @@ export class HttpClient implements ifm.IHttpClient {
...
 
-                const redirectUrl: any = response.message.headers["location"];
+                const redirectUrl: string | null = response.message.headers["location"];
                 if (!redirectUrl) {
                     // if there's no location to redirect to, we won't
                     break;
                 }
+                let parsedRedirectUrl = url.parse(redirectUrl);
+                if (parsedUrl.protocol == 'https:' && parsedUrl.protocol != parsedRedirectUrl.protocol && !this._allowRedirectDowngrade) {
+                    throw new Error("Redirect from HTTPS to HTTP protocol.");
```
damccorm: Can we make this error a little more explicit as to why this is bad and options to get around it if you're fine downgrading.

Something like `Redirect from HTTPS to HTTP protocol. This downgrade is not allowed for security reasons. If you want to allow this behavior, set the allowRedirectDowngrade option to true`

lib/HttpClient.ts
```diff
@@ -91,7 +91,8 @@ export class HttpClient implements ifm.IHttpClient {
     private _httpProxy: ifm.IProxyConfiguration;
     private _httpProxyBypassHosts: RegExp[];
     private _allowRedirects: boolean = true;
-    private _maxRedirects: number = 50
+    private _allowRedirectDowngrade: boolean = false;
```
damccorm: I want to call this out as a breaking change. This is fine IMO, since we've decided this is a security vulnerability (which I agree with to be clear), and are now considering the downgrading behavior to be a bug.

damccorm:
@bryanmacfarlane this looks good to me, but I'll defer to you on this one. Thoughts on versioning? My instinct is to make this patch if we're considering this a security bug.

lib/HttpClient.ts
JLLeitschuh: Sure!

JLLeitschuh:
> My instinct is to make this patch if we're considering this a security bug.

If this is considered a security bug, should a CVE number be assigned?

lib/HttpClient.ts
JLLeitschuh:
```suggestion
                    throw new Error("Redirect from HTTPS to HTTP protocol. This downgrade is not allowed for security reasons. If you want to allow this behavior, set the allowRedirectDowngrade option to true.");
```

test/units/httptests.ts
```diff
@@ -245,7 +245,45 @@ describe('Http Tests', function () {
...
+        nock('http://microsoft.com')
+            .get('/')
+            .reply(200, {
+                source: "nock"
+            });
+        try {
+            await _http.get('https://microsoft.com/redirect-to');
+            assert(false, "The above should fail");
+        } catch (err) {
+            assert.equal((err as Error).message, "Redirect from HTTPS to HTTP protocol.", "Error information.")
```
JLLeitschuh:
```suggestion
            assert.equal((err as Error).message, "Redirect from HTTPS to HTTP protocol. This downgrade is not allowed for security reasons. If you want to allow this behavior, set the allowRedirectDowngrade option to true.", "Error information.")
```

lib/HttpClient.ts
bryanmacfarlane: I would simply say "Downgrading from HTTPS to HTTP is not permitted on a redirect.  If this is intended, set allowRedirectDowngrade to true"

lib/HttpClient.ts
bryanmacfarlane: I think it's fine since it's calling out an odd and less secure behavior.

bryanmacfarlane:
Bump the package.json by a minor version since it's a (more secure) behavior change

bryanmacfarlane:
This is more of a security in depth code change and making it more explicit if you want that behavior.

JLLeitschuh:
Confirming: There is a general feeling here that this fix does not warrant a CVE number, correct?

damccorm:
Correct

PR #174, opened 2019-10-11
Bump minor version

damccorm:


bryanmacfarlane:
Can you remove this failing test from previous PR?: Http Tests does basic https get request with long unicode response

I was more interested that the change that went with that didn't regress any other tested behavior (and didn't necessarily want a long request test in a fast unit test set).

That will allow tests to pass and take this PR.

Then we can do a release

bryanmacfarlane:
dupe of https://github.com/microsoft/typed-rest-client/pull/175

PR #189, opened 2019-12-18
Integrate decoding gzip encoded response of Http Requests

### LINKED ISSUE #148: Fetching GZipped content is not supported
1. Please check our current Issues to see if someone already reported this https://github.com/Microsoft/typed-rest-client/issues

Done.

2. Take a look at our Instructions for logging issues https://github.com/Microsoft/typed-rest-client/blob/master/CONTRIBUTING.md#instructions-for-logging-issues

### Environment
Node version: v8.7.0
Npm version: v6.7.0
OS and version: OSX v10.13.6
typed-rest-client version: 1.4.0

### Issue Description
Making a request to a gzipped resource fails to decode the content properly.

### Expected behaviour
Gzipped endpoints can be fetched and decoded properly.

### Actual behaviour
Gzip support isn't implemented.

### Steps to reproduce
1. Add a new GET mocha test to fetching http://httpbin.org/gzip
2. Modify the test to look for headers.Host within the response ("url" is not included in /gzip endpoint).
2. Run the included tests: npm run build && npm test 
Notice the test fails.

hamzahejja:
**Description**
This PR intends to integrate the ability to decode gzipped (gzip-encoded) Http responses, depending on Node.js native _built-in_ **zlib** module.

- Added http tests to cover the above addressed case.
- Fixes #148


lib/Util.ts
```diff
@@ -75,4 +76,39 @@ function buildParamsStringifyOptions(queryParams: IRequestQueryParams): any  {
...
+ * Obtain Response's Content Charset.
+ *
+ * @param {IHttpClientResponse} response
+ * @return {string} - Content Encoding Charset; Default=utf-8
+ */
+export function obtainContentCharset (response: IHttpClientResponse) : string {
+  // Find the charset, if specified.
+  // Match with regex "charset="
+  const contentType: string = response.message.headers['content-type'] || '';
+  const matches: (RegExpMatchArray|null) = contentType.match(/charset=([^;,\r\n]+)/i);
```
damccorm: Could you maybe clarify a little bit what this regex is doing? It looks fine to me, but just for readability it might be good.

Maybe just add an example (e.g. `FOOcharset=BAR` will get matched to `[charset=BAR, BAR]`) - that also makes it clear why you're taking matches[1] below.

lib/HttpClient.ts
```diff
@@ -52,6 +53,28 @@ export class HttpClientResponse implements ifm.IHttpClientResponse {
...
+                    const chunk = (typeof data === 'string') ? Buffer.from(data, encodingCharset) : data;
+                    buffer = Buffer.concat([buffer, chunk]);
+                }).on('end', async function() {
+                    const output = await util.decompressGzippedContent(buffer, encodingCharset);
+                    resolve(output);
+                }).on('error', function(err) {
+                    reject(err);
+                });
+            });
+        }
```
damccorm: Nit: We can probably consolidate this into a single return. Would be good to get your buffer/string detection in both anyways. So the single returned promise would look something like:

```
return new Promise<string>(async (resolve, reject) => {
    let buffer: Buffer = Buffer.alloc(0);
    const encodingCharset = util.obtainContentCharset(this);

    this.message.on('data', function(data: string|Buffer) {
        const chunk = (typeof data === 'string') ? Buffer.from(data, encodingCharset) : data;
        buffer = Buffer.concat([buffer, chunk]);
    }).on('end', async function() {
        if (isGzippedEncoded) {
            resolve(await util.decompressGzippedContent(buffer, encodingCharset));
        } else {
            resolve(buffer.toString());
        }
    }).on('error', function(err) {
        reject(err);
    });
});
```

Thoughts?

lib/Util.ts
hamzahejja: Yeah, Gotcha! Will add helper comments to clarify the functionality, and add an example of matching for the charset 👍

lib/HttpClient.ts
hamzahejja: Yeah, Totally agree here too! I did think of producing a sole return statement with joining code, don't know why left it that way. Event the inner `if-else` block within the `on('end')` event can be consolidated to a sole if statement to handle gzipped content, OR ELSE return the stringified JSON buffer.

After re-arranging, sorting-out the code, turns out to be something as:
```
readBody(): Promise<string> {
        return new Promise<string>(async (resolve, reject) => {
            let buffer: Buffer = Buffer.alloc(0);
            const encodingCharset = util.obtainContentCharset(this);

            // Extract Encoding from header: 'content-encoding'
            // Match `gzip`, `gzip, deflate` variations of GZIP encoding
            const contentEncoding: string = this.message.headers['content-encoding'] || '';
            const isGzippedEncoded: boolean = new RegExp('(gzip$)|(gzip, *deflate)').test(contentEncoding);

            this.message.on('data', function(data: string|Buffer) {
                const chunk = (typeof data === 'string') ? Buffer.from(data, encodingCharset) : data;
                buffer = Buffer.concat([buffer, chunk]);
            }).on('end', async function() {
                if (isGzippedEncoded) {
                    const gunzippedBody = await util.decompressGzippedContent(buffer, encodingCharset);
                    resolve(gunzippedBody);
                }

                resolve(buffer.toString(encodingCharset));
            }).on('error', function(err) {
                reject(err);
            });
        });
    }
```

damccorm: Yep, that looks great!

strefethen:
@hamzahejja thank you. Much appreciated.

lib/Util.ts
```diff
@@ -75,4 +76,46 @@ function buildParamsStringifyOptions(queryParams: IRequestQueryParams): any  {
...
+ * @return {Promise<string>}
+ */
+export async function decompressGzippedContent(buffer: Buffer, charset?: string): Promise<string> {
+    return new Promise<string>(async (resolve, reject) => {
+        zlib.gunzip(buffer, function (error, buffer) {
+            if (error) {
+                reject(error);
+            }
+
+            resolve(buffer.toString(charset || 'utf-8'));
```
rumblefrog: Tested with a jsdelivr endpoint, zlib errors with `incorrect header check` and buffer here is undefined.

hamzahejja: I've just tested one of **jsDelivr**'s endpoints, specifically: `https://raw.githubusercontent.com/jsdelivr/jsdelivr/master/Privacy%20Policy%20-%20jsdelivr.net.md` which has GZipped (gzip encoded) content, and the string body was parsed neatly as string, with no Errors. Moreover, this PR included tests against GZipped contents endpoints, and they passed successfully. If you can, please provide the endpoint you've tried and provided the error above.

strefethen:
@hamzahejja I just upgraded to this latest version and I'm finding this code doesn't properly handle the API that I'd written my original changes for which returns a gzipped result. In the readBody method I'm seeing both resolve functions called which seems like a bug. There's an if block for isGzippedEncoded but no else for the following resolve.

Both
`resolve(gunzippedBody);`
AND
`resolve(buffer.toString(encodingCharset));`
...are getting called.

Assuming that adding an else block for the latter resolve, the gunzippedBody is subsequently lost in the fulfilled function of the __awaiter in this code
`step(generator.next(value)`
When step(result) is called the result is null and the content is lost.

I'm curious if this has been tested against a gzipped API? I'm using the following API from Postman with no problems and my prior code also works so I'm still left monkey patching this repo unfortunately.

https://api.511.org/traffic/events?format=json&api_key=<remove>&status=active&Bbox=-122.167735,37.408920,-121.911378,37.037749&version=v1

Although I've removed my API key for obvious reasons.

hamzahejja:
@strefethen As far as regarding the `readBody()` function, an if-else block should lay there for resolving the **Promise** , it's a small-slip, Thanks for the catch! As it is your catch, Would you like to raise a FIX PR ? 

>I'm curious if this has been tested against a gzipped API? I'm using the following API from Postman with no problems and my prior code also works so I'm still left monkey patching this repo unfortunately.

YES, it sure has, added tests here: https://github.com/microsoft/typed-rest-client/pull/189/files#diff-702768cf76b4d9258743dd005b0cf940R89-R106 uses this API [HttpBin](http://httpbin.org/gzip) which simulates an API with **GZIPPED ** content

strefethen:
@hamzahejja unfortunately due to this issue I've had to move away from using this package as I didn't want to keep manually patching the code. I do believe more testing is warranted as the API I mentioned above simply does not work with this code as-is and returns null.

I saw, the PR and appreciate you offering the chance for me to submit it. :)

Best!

Automobileservices4:
#189

bohonskyb:
Cannot find proof, but think in browser environment decompression handled by browser. If so `decompressGzippedContent` will throw error.

lpcarignan:
I get a `Cannot read property 'toString' of undefined` error on line 93 of Util.js. 
 
`resolve(buffer.toString(charset || 'utf-8'));`

The buffer is undefined when running Chrome 86.0.4240.198 in debug mode with the following command line:

> sudo /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --remote-debugging-port=9222 --disable-web-security --user-data-dir=$HOME/chromeTemp

If I remove `--disable-web-security --user-data-dir=$HOME/chromeTemp` from the command line, I get more cryptic error message:

`Failed to fetch `

Any ideas why the buffer would be undefined? I would propose a check on undefined before calling toString().

I'm using version typed-rest-client@1.8.0 inside the project https://github.com/microsoft/azure-devops-node-api.

YogeshIITK:
@lpcarignan,  did you find any solution for  "Cannot read property 'toString' of undefined error on line 93 of Util.js.".

YogeshIITK:
> I get a `Cannot read property 'toString' of undefined` error on line 93 of Util.js.
> 
> `resolve(buffer.toString(charset || 'utf-8'));`
> 
> The buffer is undefined when running Chrome 86.0.4240.198 in debug mode with the following command line:
> 
> > sudo /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --remote-debugging-port=9222 --disable-web-security --user-data-dir=$HOME/chromeTemp
> 
> If I remove `--disable-web-security --user-data-dir=$HOME/chromeTemp` from the command line, I get more cryptic error message:
> 
> `Failed to fetch `
> 
> Any ideas why the buffer would be undefined? I would propose a check on undefined before calling toString().
> 
> I'm using version typed-rest-client@1.8.0 inside the project https://github.com/microsoft/azure-devops-node-api.

YogeshIITK:
@Ipcarignan , did you find any solution for error "Cannot read property 'toString' of undefined error on line 93 of Util.js."

PR #191, opened 2019-12-22
[fix]: fix TypeError on socket.end() and ensure proper error message on timeouts

### LINKED ISSUE #159: Bad error message on timeouts.
When we timeout, we call `socket.end` which errors with `socket.end is not a function`

https://github.com/microsoft/typed-rest-client/blob/db388ca114dffc1e241ae81e6f3b9cd022c5b281/lib/HttpClient.ts#L339

See https://github.com/microsoft/azure-pipelines-tool-lib/pull/55 for discussion

hamzahejja:
- Use `socket.destroy()` to ensure no more I/O happens on socket( i.e. closed), in favour of `socket.end()` which currently causes a **TypeError**.
Ref. [socket.destroy([exception])](https://nodejs.org/api/net.html#net_socket_destroy_exception)
- This added logic will ensure a proper Error with a comprehensive message: _Request timeout: /path_ is returned as result of Socket timeouts.
- Added a  _nock_ unit-test to address issue.
- Fixes #159 

**Change**
_Current Behavior_: `Uncaught TypeError: socket.end is not a function`
_Added Behaviour_: `Error: Request timeout: /path` which is triggered by Promise rejection, due socket timeouts.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -370,7 +370,7 @@ export class HttpClient implements ifm.IHttpClient {
         // If we ever get disconnected, we want the socket to timeout eventually
         req.setTimeout(this._socketTimeout || 3 * 60000, () => {
             if (socket) {
-                socket.end();
+                socket.destroy();
             }
             handleResult(new Error('Request timeout: ' + info.options.path), null);
         });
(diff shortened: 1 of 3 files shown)


hamzahejja:
Unit-Test removed as `.socketDelay([delay])` of nock breaks on Node.js version 12:
- tried version `10.16.0` and worked perfectly.
- tried version `12.14.0` which is LTS of Node12 and it broke.

-- The failing builds confirm that tests fail on latest-Node 12. However, now thinking about it, the test is expendable, not needed!

PR #193, opened 2019-12-30
UserAgent should allow null/undefined

damccorm:
We already allow this in the Http Client, just getting the typing right in the restClient


brandontaylor1994:
When will this change be incorporated into a release?

stephenmichaelf:
@damccorm Can we do a release of the lib?

@brandontaylor1994 This shouldn't be blocking in any way as it's just a type check.

PR #194, opened 2020-01-02
Write nock tests for all handlers and change codebase for NTLM handler

hamzahejja:
### Description:
- Per issue #90 , This PR's purpose is to deliver nock tests for all Authentication handlers: **Basic Auth, Bearer Token, Personal Access Token (PAT), and most importantly NTLM**.
- While writing nock test for the NTLM handler, it has been observed that the current codebase produces incorrect/mismatched value(s) for type1 message, following on the [example here](https://www.innovation.ch/personal/ronald/ntlm.html), therefore a switch to [node-ntlm, NTLM authentication and Samba LM/NT hash library](https://github.com/Node-SMB/ntlm) which produces a matching values for type1, type2 and the response header (type3)
- Examples followed to verify working functionality of NTLM Authentication handler:
Bottom-of-page example [here](https://www.innovation.ch/personal/ronald/ntlm.html)
Example [here](http://davenport.sourceforge.net/ntlm.html#ntlmHttpAuthentication)

- **Nock** tests for the Authentication handlers were placed in a separate file `/tests/units/handlers.ts`.


lib/handlers/ntlm.ts
```diff
@@ -146,12 +128,30 @@ export class NtlmCredentialHandler implements ifm.IRequestHandler {
             throw new Error('www-authenticate not found on response of second request');
```
damccorm: Can you change the comment at the top of this function to point to the new library we're using?

lib/handlers/ntlm.ts
```diff
@@ -146,12 +128,30 @@ export class NtlmCredentialHandler implements ifm.IRequestHandler {
...
 
-        const type2msg = ntlm.parseType2Message(res.message.headers['www-authenticate']);
-        const type3msg = ntlm.createType3Message(type2msg, this._ntlmOptions);
+        /**
+         * Server will respond with challenge/nonce
+         * assigned to response's "WWW-AUTHENTICATE" header
+         * and should adhere to RegExp /^NTLM\s+(.+?)(,|\s+|$)/
+         */
+        const serverNonceRegex: RegExp = /^NTLM\s+(.+?)(,|\s+|$)/;
+        const serverNonce: Buffer = Buffer.from(
```
damccorm: What happens here if the server sends back a bad response? It looks like the decodeType2 function fails, we might want to catch that and throw a slightly nicer error (something like `decoding server's type 2 message failed with error: [original message here]`)

lib/opensource/Node-SMB/README.md
```diff
@@ -0,0 +1,78 @@
+# node-ntlm, NTLM authentication and Samba LM/NT hash library
```
damccorm: Can we cut this? Maybe replace it with something similar to https://github.com/microsoft/typed-rest-client/blob/master/lib/opensource/node-http-ntlm/readme.txt

lib/opensource/Node-SMB/lib/ntlm.js
```diff
@@ -0,0 +1,236 @@
+/*
```
damccorm: Can we move this copyright to https://github.com/microsoft/typed-rest-client/blob/master/LICENSE

damccorm: We should also then be able to get rid of it from other files and shrink the library a little bit

lib/handlers/ntlm.ts
```diff
@@ -6,7 +6,7 @@ import http = require("http");
 import https = require("https");
 
 const _ = require("underscore");
-const ntlm = require("../opensource/node-http-ntlm/ntlm");
```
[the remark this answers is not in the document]
hamzahejja: I did that then restored it, was bit hesitated bout' doing that, but will wipe it out 👍

lib/handlers/ntlm.ts
hamzahejja: Well, looking into [SamDecrock httpntlm module](https://github.com/SamDecrock/node-http-ntlm/blob/master/httpntlm.js#L41-L59), This function is indeed an adaptation of the original implementation, it is **NOT** of the newly used module, no ?

lib/handlers/ntlm.ts
hamzahejja: Yeah, one should intercept bad scenarios as well 👍 .. Will throw an error here to reflect such issue

lib/handlers/ntlm.ts
damccorm: Yeah, I guess you're right it still probably is. My mistake!

damccorm:
Looks like we're not going to hear back on the issue, so I'm going to merge. Thanks for doing this.

Along these lines: as we've reviewed having the ntlm handler, we've decided that now that its in a better state we probably want to draw a line and stop supporting any additional ntlm-specific changes. I'll add a PR with an explanatory comment, but basically ntlm has been flagged for some pretty sever security boundaries, there's pretty substantive literature behind that:

- https://securiteam.com/securityreviews/5op0b2kgac/
- https://www.bleepingcomputer.com/news/security/new-microsoft-ntlm-flaws-may-allow-full-domain-compromise/
- https://docs.microsoft.com/en-us/archive/blogs/miriamxyra/stop-using-lan-manager-and-ntlmv1 - this one is us (MS) telling people to stop using NTLMV1, and pointing to some issues surrounding V2 that have since been discovered to be worse than thought
- https://www.helpnetsecurity.com/2019/10/10/ntlm-vulnerabilities/

Basically, lots of people still use this so we'll keep it around, but its not our recommended mode of operation and we're going to stop further investment.

Thanks for getting this in a good place where we feel comfortable doing that!

PR #196, opened 2020-01-10
Deprecate ntlm

damccorm:
@hamzahejja FYI

### DIFF
--- a/lib/handlers/ntlm.ts
+++ b/lib/handlers/ntlm.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-// This handler has been deprecated and new enhancements beyond basic bug fixes will not be taken.
+// This handler has been deprecated.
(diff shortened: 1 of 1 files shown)


lib/handlers/ntlm.ts
```diff
@@ -1,6 +1,14 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+// This handler has been deprecated and new enhancements beyond basic bug fixes will not be taken.
```
stephenmichaelf: Nit: Maybe drop " and new enhancements beyond basic bug fixes will not be taken".

stephenmichaelf: Approving, let's get Zac to sign off too.

damccorm: @zachariahcox - are you comfortable with this? Basically, we want to deprecate the NTLM handler because its known to be insecure and we have other better ways to authenticate. We'll leave it in the library because lots of people still use NTLM, but won't take anything new.

PR #197, opened 2020-01-31
fix: Add if-else block before Resolving the Promise in HttpClient's readBody() function

hamzahejja:
- This PR fixes a bug where the `readBody()` function which returns a **Promise** has two consecutive `resolve()` of which both gets called.
- An `if-else` block is added to resolve the Promise only **once**.

### DIFF
--- a/test/tests/httptests.ts
+++ b/test/tests/httptests.ts
@@ -43,7 +43,7 @@ describe('Http Tests', function () {
         assert(res.message.statusCode == 200, "status code should be 200");
         let body: string = await res.readBody();      
         let obj: any = JSON.parse(body);
-        assert(obj.url === "https://httpbin.org/get");
+        assert(obj.url === "http://httpbin.org/get");
         assert('User-Agent' in obj.headers === true, "User-Agent should be set");
     });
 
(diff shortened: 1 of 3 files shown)


test/tests/httptests.ts
```diff
@@ -43,7 +43,7 @@ describe('Http Tests', function () {
         assert(res.message.statusCode == 200, "status code should be 200");
         let body: string = await res.readBody();      
         let obj: any = JSON.parse(body);
-        assert(obj.url === "https://httpbin.org/get");
+        assert(obj.url === "http://httpbin.org/get");
```
damccorm: Any reason for these test changes?

hamzahejja: @damccorm For some reason, the tests were failing! Assertions on `obj.url` were not fulfilled as expected.  You could check/see this in the first commit where the builds failed as tests failed. I wonder why in the first place tests were the `https` module is not used, is asserting against urls with `https://`. But still I'm confused how they working fine previously and out-of-sudden first when opening this PR

damccorm: Hm, it looks like the payload just changed http://httpbin.org/get

Thanks for clarifying, this LGTM!

PR #200, opened 2020-02-17
Makes RestClient._processResponse protected

### LINKED ISSUE #199: Make RestClient._processResponse protected instead of private
typed-rest-client version: 1.7.2

I'd like 404s to throw, as I'm sure many other consumers of this library would.  This would be simple to do in a derived class if we could override _processResponse.  Furthermore, this would allow derived classes to implement additional methods while leveraging the existing response processing.

AnthonyDiSanti:
Resolves #199

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -60,7 +60,7 @@ export class RestClient {
         let url: string = util.getUrl(requestUrl, this._baseUrl);
         let res: httpm.HttpClientResponse = await this.client.options(url,
             this._headersFromOptions(options));
-        return this._processResponse<T>(res, options);
+        return this.processResponse<T>(res, options);
     }
 
     /**
@@ -75,7 +75,7 @@ export class RestClient {
         let url: string = util.getUrl(resource, this._baseUrl, (options || {}).queryParameters);
         let res: httpm.HttpClientResponse = await this.client.get(url,
             this._headersFromOptions(options));
-        return this._processResponse<T>(res, options);
+        return this.processResponse<T>(res, options);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Thanks for contributing to the lib, @AnthonyDiSanti ! Generally what consumers do here is wrap the code and throw themselves on 404 if they need to. We would prefer to not expose this method if we can avoid it.

Would that work for you?

AnthonyDiSanti:
That's the approach I'm taking today, but it requires wrapping each public method.  I find that reflection tends to throw off the vscode typescript parser, reducing its usefulness, so I'm maintaining the list of public methods in my derived class.  The alternative implementation would be include this 404 boilerplate everywhere I use the class.  In either case, wrapping _processResponse directly is a cleaner solution.

I understand why you wouldn't want to make this method public, but derived classes would greatly benefit from access to it.  If the team has strong feelings that this method should not be exposed, no problem, I very much appreciate your work maintaining this and I'll continue using the workaround.  However, I really do think this would be an improvement.

stephenmichaelf:
@AnthonyDiSanti We had another discussion on this and we are OK making the change. Would you mind removing the `_` at the start of the function since it won't be private?

Thanks for taking the time to provide thoughtful responses!

AnthonyDiSanti:
Thank you for reconsidering and approving this request.  I've renamed `_processRequest` to `processRequest` as per your guidance, squashed the commits, and rebased off the latest master.  I've also confirmed that all tests pass locally for me and signed the CLA.  Please let me know if there's anything else I can do to help.  Thank you again!

PR #207, opened 2020-03-31
Disable authentications on redirections

yahavi:
This PR fixes a security issue.

**Current behavior**
1. Send any request with `BasicCredentialHandler`, `BearerCredentialHandler` or `PersonalAccessTokenCredentialHandler`.
2. The target host may return a redirection (3xx), with a link to a second host.
3. The next request will use the credentials to authenticate with the second host, by setting the `Authorization` header.

**Expected behavior**
3. The next request will NOT set the `Authorization` header.

See the same issue for CURL: 
1. [HTTP authentication leak in redirects](https://curl.haxx.se/docs/CVE-2018-1000007.html) - I used the same solution as CURL did.
2. [CVE-2018-1000007](https://nvd.nist.gov/vuln/detail/CVE-2018-1000007).

### DIFF
--- a/lib/handlers/basiccreds.ts
+++ b/lib/handlers/basiccreds.ts
@@ -6,20 +6,24 @@ import ifm = require('../Interfaces');
 export class BasicCredentialHandler implements ifm.IRequestHandler {
     username: string;
     password: string;
-    host: string;
+    allowCrossOriginAuthentication: boolean;
+    origin: string;
 
-    constructor(username: string, password: string) {
+    constructor(username: string, password: string, allowCrossOriginAuthentication?: boolean) {
         this.username = username;
         this.password = password;
+        this.allowCrossOriginAuthentication = allowCrossOriginAuthentication;
     }
 
     // currently implements pre-authorization
     // TODO: support preAuth = false where it hooks on 401
     prepareRequest(options:any): void {
+        if (!this.origin) {
+            this.origin = options.host;
+        }
         // If this is a redirection, don't set the Authorization header
-        if (!this.host || this.host === options.host) {
+        if (this.origin === options.host || this.allowCrossOriginAuthentication) {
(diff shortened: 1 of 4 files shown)


damccorm:
I don't think we should introduce this change without a way to override it since this is a breaking change, but I generally agree with the change.

Could you add another parameter to PrepareRequest to control this (and propogate it back where it gets called in httpclient)? It can probably just be an extra requestOption in the constructor

yahavi:
Thanks for the response. 
I agree this can be configurable.
What do you think on matching on [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) request header?

**3 Options:**
1. `Access-Control-Allow-Origin` is falsy - Don't allow authorization with other hosts. (This is the default)
2. `Access-Control-Allow-Origin=*` - Allow authorization with all hosts.
3. `Access-Control-Allow-Origin=<host-name>` - Allow authorization with the the first host AND with `<host-name>`.

yahavi:
After more thoughts, I concluded that from user's perspective it'll be much simpler to add an optional boolean `allowCrossOriginAuthentication` to the constructor.

Changes in the last commit:
* Add `allowCrossOriginAuthentication?: boolean` to the constructor.
* Change `host` -> `origin` to be more specific.

JLLeitschuh:
Was a CVE ever issued for this vulnerability?

JLLeitschuh:
I've reported this upstream to MSRC to get a CVE issued.
https://msrc.microsoft.com/report-detail/VULN-047585

JLLeitschuh:
I've requested a CVE number from MITRE as I received the following from Microsoft:

> CVEs are generally only released from Microsoft for very specific issues.  Specifically they are used for items that require patching through Windows Update.  They are used to explain why a certain patch is needed and are tied into that download.   If a fix is to an online service or online repo like GitHub locations, these would not receive a CVE.  This item does not look like it was tied into the Windows Update release and would not be eligible for a CVE.

PR #210, opened 2020-04-07
Fix security warnings

damccorm:
Right now we're getting flagged for a bunch of security warnings. This fixes some of them

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,15 +17,6 @@ steps:
 - script: npm run units
   displayName: npm run units
 
-# test using node 4.8.6
-- task: NodeTool@0
-  inputs:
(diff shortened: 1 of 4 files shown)


stephenmichaelf:
Did you manually change the package-lock?

damccorm:
I used `npm audit --fix` - npm automatically manages these. It didn't fix all of the issues because of incompatibilities, but these ones didn't cause any issues

stephenmichaelf:
Is that going to break when we make future updates to package versions?

damccorm:
No, the package-lock won't change unless a new package version is introduced that requires it to. So not unless we explicitly introduce a dependency on a lower version of one of these things

PR #211, opened 2020-04-11
Remove deprecated URL API usage

JamieMagee:
The legacy URL API (provided by `require('url')`) was deprecated by NodeJS 11 released in October 2018. It is scheduled to be removed in a future release.

The main motivation for this change, is Yarn v2, which does not allow undeclared dependencies to be used. `require('url')` is ambiguous between the built-in NodeJS module, and the npm package [url](https://www.npmjs.com/package/url). I'm assuming the intent is to use the built-in module, so I've migrated to using the new URL API, for which the import is implicit.

### DIFF
--- a/lib/Util.ts
+++ b/lib/Util.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 import * as qs from 'qs';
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Closing due to inactivity.

PR #212, opened 2020-04-11
Remove ntlm files from repo

JamieMagee:
Instead, add git dependency in package.json


damccorm:
For security reasons we can't do this from a non-trusted source, see this comment - https://github.com/microsoft/typed-rest-client/issues/90#issuecomment-570293148

JamieMagee:
Okay, but what's the difference between this, and the [other dependencies from package.json?](https://github.com/microsoft/typed-rest-client/blob/master/package.json#L42-L45). Would it make any difference if this were tagged against a specific git hash?

damccorm:
> Okay, but what's the difference between this, and the other dependencies from package.json?.

Essentially level of comfort with the publisher. We've got an approved list on our side that we're comfortable taking a dependency on.

> Would it make any difference if this were tagged against a specific git hash?

Not really - this is mostly a compliance issue (not necessarily even stemming from Microsoft, some of these requirements come from downstream customers)

PR #216, opened 2020-05-28
Fix contrstructor on NO_PROXY with wildcard domain

### LINKED ISSUE #215: Environment variable "NO_PROXY" with wildcard causes HttpClient constructor failure
### Environment
Node version: 10.13.0
Npm version: 6.4.1
OS and version: Windows Server 2012 R2
typed-rest-client version: 1.7.3

### Issue Description
Some environments define the environment variable `NO_PROXY` to bypass proxy servers, which the `HttpClient` of this package respects. It is allowed to set wildcard domains in this environment variable like `*.microsoft.com`, altough not every client must respect them. However the `HttpClient` constructor crashes when environment variables like these are present.

### Expected behaviour
The `HttpClient` should construct normally.

### Actual behaviour
The `HttpClient` constructor throws an Exception:
Invalid regular expression: /*microsoft.com/: Nothing to repeat

### Steps to reproduce
1. Set an environment variable like `NO_PROXY = microsoft.com,*.microsoft.com`
2. Construct a `HttpClient`

### Additional Info
This problem caused our Azure DevOps Release Pipeline to crash which tried to deploy an App Service to Azure using the `AzureRmWebAppDeploymentV4` Task. The task already crashes when importing its (js) dependencies, so it was a little complicated to find the source of this error to be this package 😉

tuwrraphael:
This fixes #215 by escaping the wildcard for the `RegExp`. I created the utility function `buildProxyBypassRegexFromEnv` to allow unit testing of this functionality.
A second test ensures the `HttpClient` constructs with the `NO_PROXY` environment variable set to a wildcard domain.


HonzaMac (approved):

HonzaMac:
Can you please take a look on this fix?
This break underlying clients when proxy has wildcards "*".
cc @mastrzyz @damccorm 🙏

HonzaMac:
cc @bryanmacfarlane 🙏

lib/Util.ts
```diff
@@ -98,6 +98,19 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
+/**
+ * Builds a RegExp to test urls against for deciding
+ * wether to bypass proxy from an entry of the
+ * environment variable setting NO_PROXY
+ *
+ * @param {string} bypass
+ * @return {RegExp}
+ */
+export function buildProxyBypassRegexFromEnv(bypass : string) : RegExp {
+    let wildcardEscaped = bypass.replace('*', '(.*)');
```
damccorm, line 110: I'm hesitant to do this simply because this behavior has been around for years, and we risk breaking people who have already done something to work around this. I generally think its the right behavior though - I might prefer something like:

```
try {
   // We need to keep this around for back-compat purposes
   return new RegExp(bypass, 'i');
catch (<Exception type (I don't know offhand what this is)>) {
   let wildcardEscaped = bypass.replace('*', '(.*)');
   return new RegExp(wildcardEscaped, 'i');
}

tuwrraphael: I understand what you mean. I changed my implementation to escape the wildcard at the start only if it causes this problem, as suggested.
Unfourtunately it seems that the `https://httpbin.org/redirect-to` endpoint is broken, which causes integration tests to fail.

damccorm: If you pull master into your branch it should be fixed (as of 30 seconds ago 😄)

tuwrraphael: done 👍

tuwrraphael:
done 👍  Thanks for your time & support!

HonzaMac:
Thank you 🎉🎉🎉🎉

PR #222, opened 2020-06-29
Added retry handling when we have a network Error

mastrzyz:
On any http/https library request error we reject the promise in requestRaw. 

There might be a TCP network error which has occurred which we will want to retry.

I have selected the following errors to be retryable : 'ECONNRESET', 'ENOTFOUND', 'ESOCKETTIMEDOUT', 'ETIMEDOUT', 'ECONNREFUSED'


mastrzyz:
Those failed UT's , I'm also getting them on Master, is there any known issue?

PR #223, opened 2020-06-29
Added handling for when we have network failures in the retry mechanism

mastrzyz:
This will add handling when we have a network failure.
Occurs when the ADO Connection itself goes down and not return any status code.

The following network errors will be retried `'ECONNRESET', 'ENOTFOUND', 'ESOCKETTIMEDOUT', 'ETIMEDOUT', 'ECONNREFUSED'`

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -248,7 +248,7 @@ export class HttpClient implements ifm.IHttpClient {
                 response = await this.requestRaw(info, data);
             }
             catch (err) {
-                if(err && err.code && NetworkRetryErrors.indexOf(err.code) !== 1){
+                if(err && err.code && NetworkRetryErrors.indexOf(err.code) > -1){
                     continue;
                 }
                 throw err;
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typed-rest-client",
-  "version": "1.7.3",
+  "version": "1.8.3",
```
damccorm, line 3: Nit: should be `1.8.0`. Also, please run npm install so that its reflected in the package-lock.json as well

mastrzyz: Done thank you!

lib/HttpClient.ts
```diff
@@ -243,8 +244,15 @@ export class HttpClient implements ifm.IHttpClient {
 
         let response: HttpClientResponse;
         while (numTries < maxTries) {
-            response = await this.requestRaw(info, data);
-
+            try{
+                response = await this.requestRaw(info, data);
+            }
+            catch (err) {
+                if(err && err.code && NetworkRetryErrors.indexOf(err.code) !== 1){
```
damccorm, line 251: Whoops, missed this in the initial review - this should be `if(err && err.code && NetworkRetryErrors.indexOf(err.code) > -1)`. As it is currently written it will retry for anything that isn't an ENOTFOUND

mastrzyz: Sorry about that

Issue #236, opened 2020-10-30
HTTP Tests currently failing because of httpbin bug


parkerault:
### Issue Description

httpbin is currently returning 404 on all requests to `redirect-to`. This issue: https://github.com/postmanlabs/httpbin/issues/617 has been open since June 20th, 2020. Most of the mentioned PRs on that issue are migrating to httpbingo, but it might be better to spin up a local server to remove external test dependencies.

PR #238, opened 2020-11-06
Passed timeout to info.options

anatolybolshakov:
It seems that timeout is not being passed to request info - fixed it and passed it to request info also.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -418,7 +418,8 @@ export class HttpClient implements ifm.IHttpClient {
         info.options.port = info.parsedUrl.port ? parseInt(info.parsedUrl.port) : defaultPort;
         info.options.path = (info.parsedUrl.pathname || '') + (info.parsedUrl.search || '');
         info.options.method = method;
-        info.options.timeout = this.requestOptions && this.requestOptions.socketTimeout;
+        info.options.timeout = (this.requestOptions && this.requestOptions.socketTimeout) || this._socketTimeout;
+        this._socketTimeout = info.options.timeout;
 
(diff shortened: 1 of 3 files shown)


lib/HttpClient.ts
```diff
@@ -418,6 +418,7 @@ export class HttpClient implements ifm.IHttpClient {
         info.options.port = info.parsedUrl.port ? parseInt(info.parsedUrl.port) : defaultPort;
         info.options.path = (info.parsedUrl.pathname || '') + (info.parsedUrl.search || '');
         info.options.method = method;
+        info.options.timeout = this.requestOptions && this.requestOptions.socketTimeout;
```
damccorm, line 421: I think we want to also check (and set) `this._socketTimeout`. I think it should look something like

```
info.options.timeout = (this.requestOptions && this.requestOptions.socketTimeout) || this._socketTimeout;
this._socketTimeout = info.options.timeout;
```

damccorm (approved):

damccorm:
Test failures here look unrelated to the change, I think httpbin is having some issues https://github.com/microsoft/typed-rest-client/issues/236

damccorm:
I'm comfortable merging without tests passing here because of that - @anatolybolshakov I realized you didn't have write permissions, I just added the team so you should have them now

anatolybolshakov:
> I'm comfortable merging without tests passing here because of that - @anatolybolshakov I realized you didn't have write permissions, I just added the team so you should have them now

Ok, thanks! So I'm merging it currently

damccorm:
Published!

Issue #241, opened 2020-11-23
Feature request: Handle response decoding (JSON.parse) in a handler so you can provide your own


air2:
### Issue Description
I have to a rest API (well some company provides a rest API and I have to talk to it) which has some long history so some calls do not send back json, but send back XML. Which screws the parsing. If I only could provide my own `body->json-object` handler it would be great. the default handler could implement the same logic as is in it right now.

Is this a direction this project would take? If so I would maybe able to supply some code to implement this.

air2:
This is not stale

EpicKris:
Bump.

PR #242, opened 2020-11-23
Added support for content-handlers.

air2:
See #241 
With this code it is possible to include your own content parser based on the response. (for example execute your own response parser when the content-type is not application/json)
Also it is possible to include a content parser for a specific request. When no content parsers can handle the content, the original json parser is used.


air2:
I fixed the tests by using https://httpbingo.org instead of https://httpbin.org for the redirect tests.

PR #245, opened 2021-01-25
Add support for query parameters for POST, PATCH, and PUT requests

ajuanjojjj:
I noticed that this was only supported for GET and DELETE requests. Given that some APIs do actually support having query parameters for these types, I figured that adding support for them, given that it doesn't conflict with normal operation only makes the client more versatile

### DIFF
--- a/lib/RestClient.ts
+++ b/lib/RestClient.ts
@@ -104,7 +104,7 @@ export class RestClient {
         resources: any,
         options?: IRequestOptions): Promise<IRestResponse<T>> {
 
-        let url: string = util.getUrl(resource, this._baseUrl);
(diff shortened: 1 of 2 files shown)


ajuanjojjj:
Closing till i fix the tests erroring

ajuanjojjj:
Updated the test units to take into account the fact that these functions already take 2 parameters.

The Http request with redirects fails already fail on the current version on the repo.

ajuanjojjj:
Either the pull from the original or whatever domain the test was pointing fixed it, so if someone could take a look at this, it'd be very appreciated :)

ajuanjojjj:
Fixes #257

PR #286, opened 2021-08-03
Fix processing of proxybypass variables from environment

EzzhevNikita:
This PR should address the issue with the processing of proxy bypass variables. Current logic incorrectly processes any host that is passed to NO_PROXY env variable in the format: _.somehost.net_. Such hosts will be wrongly interpreted because they contain points in hosts. Point is a special symbol for regex, so it should be escaped by backslash.


lib/Util.ts
```diff
@@ -107,17 +107,17 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
-        // We need to keep this around for back-compat purposes
-        return new RegExp(bypass, 'i')    
-    }
-    catch(err) {
-        if (err instanceof SyntaxError && (bypass || "").startsWith("*")) {            
-            let wildcardEscaped = bypass.replace('*', '(.*)');
-            return new RegExp(wildcardEscaped, 'i');
-        }
-        throw err;
+    let searchRegExpToReplaceSpecialChars = new RegExp('([.])', 'g');
```
alexander-smolyakov, line 110:
``` suggestion
    const searchRegExpToReplaceSpecialChars: RegExp = new RegExp('([.])', 'g');
```

lib/Util.ts
```diff
@@ -107,17 +107,17 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
+
+    // replace all . symbols in string by \. because point is a special character
+    let safeRegex = (bypass || "").replace(searchRegExpToReplaceSpecialChars, '\\$1');
```
alexander-smolyakov, line 113:
```suggestion
    let safeRegex: string = (bypass || "").replace(searchRegExpToReplaceSpecialChars, '\\$1');
```

lib/Util.ts
```diff
@@ -107,17 +107,17 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
+    
+    // check if expression starts with asterisk and replace it with .*
+    if (safeRegex && safeRegex.startsWith("*")) {
+        safeRegex = safeRegex.replace("*", ".*");
     }
+
+    return new RegExp(safeRegex, 'i');
```
alexander-smolyakov, line 120: Could it make sense to expand our test coverage for this function?

https://github.com/microsoft/typed-rest-client/blob/9397d2e164afe5f70119dde75935dba9d09d7afd/test/units/utiltests.ts#L17-L44

EzzhevNikita: Yeah, make sense

EzzhevNikita: Added several tests that cover cases that I am trying to fix by this PR

alexander-smolyakov (approved):
LGTM, thanks!

lib/Util.ts
```diff
@@ -107,17 +107,17 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
-    catch(err) {
-        if (err instanceof SyntaxError && (bypass || "").startsWith("*")) {            
-            let wildcardEscaped = bypass.replace('*', '(.*)');
-            return new RegExp(wildcardEscaped, 'i');
-        }
-        throw err;
+    const searchRegExpToReplaceSpecialChars: RegExp = new RegExp('([.])', 'g');
+
+    // replace all . symbols in string by \. because point is a special character
+    let safeRegex: string = (bypass || "").replace(searchRegExpToReplaceSpecialChars, '\\$1');
```
anatolybolshakov, line 113: Won't it break customers which use regexp as a value with dots as special characters?

EzzhevNikita: Generally NO_PROXY env variable shouldn't contain a regex inside, the strings the set inside should be interpreted as is. The reason why we need to escape dot symbols is that we look for the specific domains in URLs using the RegExp test function which accepts a regular expression as a pattern to search, so we transform the string to the RegExp. NO_PROXY is quite a common way of setup a proxy bypass list and as far as I see there is no soft that supports regexp inside, so dots shouldn't be used as special characters there and in my mind, they could be safely escaped.

EzzhevNikita: And also our docs related to the setup of proxy bypass domains is saying that dots should be escaped - https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/proxy?view=azure-devops&tabs=windows#specify-proxy-bypass-urls

lib/Util.ts
```diff
@@ -107,17 +107,17 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
-        throw err;
+    const searchRegExpToReplaceSpecialChars: RegExp = new RegExp('(?<!\\\\)([.])(?!\\*)', 'g');
+
+    // check if expression starts with asterisk and replace it with .*
+    if (bypass && bypass.startsWith("*")) {
+        bypass = bypass.replace("*", ".*");
     }
+
+    // replace all . symbols in string by \. because point is a special character
+    var safeRegex = (bypass || "").replace(searchRegExpToReplaceSpecialChars, '\\$1');
```
alexander-smolyakov, line 118: Nit: Use `let` or `const` please

alexander-smolyakov (approved):
LGTM and please take a look at the comment

lib/Util.ts
EzzhevNikita: done

lib/Util.ts
```diff
@@ -107,17 +107,17 @@ export async function decompressGzippedContent(buffer: Buffer, charset?: string)
...
-        // We need to keep this around for back-compat purposes
-        return new RegExp(bypass, 'i')    
-    }
-    catch(err) {
-        if (err instanceof SyntaxError && (bypass || "").startsWith("*")) {            
-            let wildcardEscaped = bypass.replace('*', '(.*)');
-            return new RegExp(wildcardEscaped, 'i');
-        }
-        throw err;
+    const searchRegExpToReplaceSpecialChars: RegExp = new RegExp('(?<!\\\\)([.])(?!\\*)', 'g');
```
anatolybolshakov, line 110: Nit: could it make sense to move it to constants and not initialize on every call?

anatolybolshakov (approved):

lib/Util.ts
EzzhevNikita: done

PR #296, opened 2021-09-07
Add node 6, 10 and 14 to GitHub Actions and Azure Pipeline tests

kuleshovilya:
Add node 6, 10 and 14 to GitHub Actions and Azure Pipeline tests, lower npm requirements in make.js due to it being unable to use node 6

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,6 +35,24 @@ steps:
 - script: npm run units
   displayName: npm run units
 
+# test using node 10.24.1
+- task: NodeTool@0
+  inputs:
+    versionSpec: "10.24.1"
+  displayName: Install node 10.24.1
+
+- script: npm run units
+  displayName: npm run units
+
+# test using node 14.17.6
+- task: NodeTool@0
+  inputs:
+    versionSpec: "14.17.6"
+  displayName: Install node 14.17.6
+
+- script: npm run units
+  displayName: npm run units
+
 # publish build artifacts to drop folder so we can use them for release
 - task: PublishBuildArtifacts@1
   inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,11 +9,10 @@ strategy:
     'Node 14': 
       versionSpec: '14.x'
 steps:
-# use node 8.9.1 this is the recommended version for building the typed rest client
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -1,8 +1,18 @@
+strategy:
+  matrix:
+    'Node 8': 
+      versionSpec: '8.x'
```
anatolybolshakov, line 4: Please add tests for node 6 also - since currently it is still used as execution handler

anatolybolshakov (approved):

DaniilShmelev (commented):
Just curious - why do we have both github actions and azure pipelines CI? As I can see, github actions runs `npm run build` and `npm test`, and azure pipelines runs `npm run units`. Should we merge them/get rid of one of them?
I think it would be easiest to just add `npm run units` command to github actions and remove azure pipelines.
@anatolybolshakov what do you think?

alexander-smolyakov (approved):

alexander-smolyakov:
> Just curious - why do we have both github actions and azure pipelines CI? As I can see, github actions runs `npm run build` and `npm test`, and azure pipelines runs `npm run units`. Should we merge them/get rid of one of them?
> I think it would be easiest to just add `npm run units` command to github actions and remove azure pipelines.
> @anatolybolshakov what do you think?

@DaniilShmelev, the azure pipeline also publishes the build artifact that we could use later to publish the package to npm.
https://github.com/microsoft/typed-rest-client/blob/8bdfb968c1207ce755685124c2139a7febf7cbad/azure-pipelines.yml#L38-L44

Overall I agree with the idea to merge these pipelines into one and if we will decide to do this let's save the pipeline in Azure.

kuleshovilya:
@alexander-smolyakov @DaniilShmelev So, what are we deciding on? Merging both tasks, as in adding unit tests and publish build to a github actions and removing azure pipelines, while saving it to azure repo?

alexander-smolyakov:
@kuleshovilya I would like to suggest the following approach:
1) Merge this PR
2) Update related Azure pipeline and remove GitHub action
3) Open another PR

DaniilShmelev (approved):

anatolybolshakov (approved):

PR #301, opened 2021-10-07
Add support for query parameters for POST, PATCH, and PUT requests

### LINKED ISSUE #257: Support for query parameters in `create`, `update` and `replace` calls
### Issue Description

I can see support for query parameters were added to the RestClient as per issue #172 however they were only added to some calls but not all. Is there a reason for that? It seems misleading that the library accepts query parameters for other calls however simply drops them without error or warning. 

As an example, when trying to query `googleapi` with a post request, it requires the `apikey` to be supplied as a query parameter. This currently seems to not be possible through the RestClient unless I specify the parameters as part of the URL.

### Expected behaviour

Query parameters should not be dropped from any request or the compiler should complain if they are not supported.

### Actual behaviour

`create`, `update` and `replace` requests do not honor the query parameters however they accept it without compilation errors.

### Steps to reproduce

1. Make the following request

```
const client = new rm.RestClient('', 'https://test.com')
const options = {
  queryParameters: {
    lang: "en",
    terms: ["foo bar", 4.2, "baz"]
  }
}
const response = await client.create('path', null, options)
```
2. Notice that the query parameters are missing from the request (however you can introspect it)

ajuanjojjj:
I noticed that this was only supported for GET and DELETE requests. Given that some APIs do actually support having query parameters for these types, I figured that adding support for them (Given that the update has no conflicting changes) only makes the client more versatile. 

My last PR had some problems with some test units failing because of a down link, but this one should be all good.

Fixes #257 and has no breaking changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,17 @@
 {
   "name": "typed-rest-client",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "typed-rest-client",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "license": "MIT",
       "dependencies": {
         "des.js": "^1.1.0",
         "js-md4": "^0.3.2",
-        "qs": "^6.9.7",
+        "qs": "^6.10.3",
         "tunnel": "0.0.6",
         "underscore": "^1.12.1"
       },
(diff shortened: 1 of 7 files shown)


ajuanjojjj:
And `https://httpbingo.org/redirect-to` seems to be failing because `certificate has expired`, although seems to be working fine when testing locally. 
Is this a known thing? If not, should I change the httpbingo tests to a different one?

ajuanjojjj:
Closing to retrigger checks

PR #316, opened 2022-03-02
Fix broken azure pipeline

DaniilShmelev:
Currently the pipeline is not working because the default image is assumed to be ubuntu-16, which is removed.
This PR changes the image to ubuntu-20
Also, this introduces a temporary fix for [node 6 / node 8 certificate issue](https://stackoverflow.com/questions/62128611/nodejs-certificate-has-expired) in unit tests.

### DIFF
--- a/test/tests/httptests.ts
+++ b/test/tests/httptests.ts
@@ -9,8 +9,9 @@ import * as path from 'path';
 
 let sampleFilePath: string = path.join(__dirname, 'testoutput.txt');
 
-if (process.version.startsWith('v8')) {
-    console.log('Setting NODE_TLS_REJECT_UNAUTHORIZED = 0 to avoid certificate errors on node 8');
+const nodeVersionsWithCertificateErrors = [6, 8];
+if (nodeVersionsWithCertificateErrors.find((nodeVersion) => process.version.startsWith(`v${nodeVersion}.`))) {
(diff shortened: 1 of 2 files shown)


alexander-smolyakov (approved):

denis-tikhomirov (approved):

test/tests/httptests.ts
```diff
@@ -9,6 +9,12 @@ import * as path from 'path';
 
 let sampleFilePath: string = path.join(__dirname, 'testoutput.txt');
 
+const nodeVersionsWithCertificateErrors = [6, 8];
+if (nodeVersionsWithCertificateErrors.find((nodeVersion) => process.version.startsWith(`v${nodeVersion}.`))) {
+    console.log('Setting NODE_TLS_REJECT_UNAUTHORIZED = 0 to avoid certificate errors on this node version');
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
AndreyIvanov42, line 15: This changes is highly insecure.

DaniilShmelev: Fixed it - the pipeline now uses the HTTP protocol instead of HTTPS in the redirect tests on node 6 / 8. This also fixes the issue.

AndreyIvanov42 (approved):

mmrazik (approved):

PR #322, opened 2022-04-19
Reduce Buffer copies in HttpClient

### LINKED ISSUE #321: Performance issues with readBody in HttpClient.ts
### Environment
Node version: 14.18.3
Npm version: 6.14.5
OS and version: Mac 10.15.7
typed-rest-client version: ^1.8.4

### Issue Description
Buffer.concat being called on repeat inside `readBody` in `HttpClient.ts` is causing electron UI to freeze.
NOTE: we do have a incomming PR on the way, we collect the `chunks` into an array and build the `Buffer` with a single concat in the 'end' phase of the `readBody()` this has been tested and proved to significantly improve performance!

### Expected behaviour
Electron client should still be quick. 

### Actual behaviour
Electron client freezes and high CPU usage while calling an api on loop

### Steps to reproduce
1. Have electron (version shouldn't matter)
2. use npm package `azure-devops-node-api` to call `GitApi` `getRepositories` (it uses typed-rest-client to call its rest api)
3. we found the issue by having 1k+ projects for an azure account then looping through getRepositories to fetch them all.
```
        return Promise.all(
          fp.map((project) => {
            return _gitApi.getRepositories(project.id, includeLinks, includeAllUrls);
          }, projects)
        );
```
4. We found it to be significantly inefficient and dug deeper to find Buffer.concat was killing performance

### Logs
profiler logs:
image
profiler logs zoomed in:
image

Mr-Wallet:
Fixes #321

Currently there are no performance tests that I could find, not sure how you folks might want to go about doing that in order to have one "fail in the absence of [my] non-test code changes." Consequently I haven't modified the tests.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -55,7 +55,7 @@ export class HttpClientResponse implements ifm.IHttpClientResponse {
     public message: http.IncomingMessage;
     readBody(): Promise<string> {
         return new Promise<string>(async (resolve, reject) => {
-            let buffer: Buffer = Buffer.alloc(0);
+            const chunks = [];
             const encodingCharset = util.obtainContentCharset(this);
 
             // Extract Encoding from header: 'content-encoding'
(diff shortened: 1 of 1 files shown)


AndreyIvanov42:
Thanks for contributing!

AndreyIvanov42 (approved):

mmrazik (approved):

anatolybolshakov (approved):

Issue #327, opened 2022-05-12
error from zlib not handled properly in decompressGzippedContent()


akrantz:
1. Please check our current Issues to see if someone already reported this https://github.com/Microsoft/typed-rest-client/issues
2. Take a look at our Instructions for logging issues https://github.com/Microsoft/typed-rest-client/blob/master/CONTRIBUTING.md#instructions-for-logging-issues

### Environment
Node version: 16.15.0
Npm version: 8.10.0
OS and version: Windows 11
typed-rest-client version: 1.8.6
browser: Edge WebView2

### Issue Description
zlib error is not handled properly in decompressGzippedContent(). "buffer" is undefined, so after if (error) reject(), it should not execute resolve(buffer.toString()).

### Expected behaviour
Should not have unhandled exception.

### Actual behaviour
Uncaught TypeError: Cannot read properties of undefined (reading 'toString')

### Steps to reproduce
![image](https://user-images.githubusercontent.com/741153/168155769-933738d4-ac25-46b2-aa8f-c1f3faf404e4.png)

akrantz:
The resolve() statement should be in the else clause if the if before it.

kirill-ivlev:
@akrantz thank you for reporting!

resolve statement should be in the else block, or we should have a return statement in the error block.

I've prepared PR which resolves this issue.

akrantz:
I believe this was fixed.

PR #328, opened 2022-05-17
fix error handling for decompressGzippedContent #327

kirill-ivlev:
**link to related issue**: #327 

**description**: 
Promise rejection wrapped on the single if without else clause will not stop method execution in case of the error, any other statements which go after the if block will be also executed, which confuses some of IDEs and may cause error in the future


anatolybolshakov (approved):

alexander-smolyakov (approved):

akrantz (approved):
Thanks for making the fix.

alexander-smolyakov:
@mmrazik could you please review this PR?

mmrazik (approved):

PR #380, opened 2024-11-20
Support for AbortController

oliver-s-lee:
Hi there, thanks for the nice project.

This small PR adds support for Node's AbortController, which is the currently recommended way to abort requests in flight. I couldn't see any way to achieve this in the current version, and as the change is relatively minor I've put this PR quickly together to demonstrate what I mean. AbortControllers are already supported by the underlying `http.request()` and `https.request()` methods; this PR simply passes an optional AbortSignal object through the various function calls to the underlying request.

Usage:

```
// Abort controllers are setup in the usual way.
const controller = new AbortController();

// RestClient also supported.
const client = new HttpClient(
    // Other options.
);

// When you make a request, you can pass the signal from the AbortController.
const request = client.get(
    "https://github.com/",
    {},  // Headers.
    controller.signal
);

// If we change our minds, we can now cancel the request.
controller.abort("No longer needed");

// Causing the promise to reject.
try {
    const val = await request;
} catch (err) {
    console.log(err); // AbortError: No longer needed
}
```

Abort controllers are (sadly) single use only  so specifying at the constructor/object level doesn't really make sense, hence the addition to the method call directly. Some of the method signatures are now getting a bit messy however, it might be more desirable to add `signal` to the `IRequestOptions` or similar instead... 

Apologies if this functionality is unwanted/covered elsewhere.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -202,36 +202,36 @@ export class HttpClient implements ifm.IHttpClient {
         }
     }
 
-    public options(requestUrl: string, additionalHeaders?: ifm.IHeaders): Promise<ifm.IHttpClientResponse> {
-        return this.request('OPTIONS', requestUrl, null, additionalHeaders || {});
+    public options(requestUrl: string, additionalHeaders?: ifm.IHeaders, signal?: AbortSignal): Promise<ifm.IHttpClientResponse> {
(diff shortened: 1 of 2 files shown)


oliver-s-lee:
@microsoft-github-policy-service agree

PR #389, opened 2025-12-24
Remove Deprecated use of url.parse 

rajmishra1997:
Deprecation in Node 24: https://nodejs.org/api/deprecations.html#DEP0169

This PR updates use of url.parse to Safer `New URL()`
Impacts how request object is formed by rest-client

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -84,13 +84,17 @@ export class HttpClientResponse implements ifm.IHttpClientResponse {
 
 export interface RequestInfo {
     options: http.RequestOptions;
-    parsedUrl: url.Url;
+    parsedUrl: URL;
     httpModule: any;
 }
 
 export function isHttps(requestUrl: string) {
-    let parsedUrl: url.Url = url.parse(requestUrl);
-    return parsedUrl.protocol === 'https:';
+    try {
+        let parsedUrl: URL = new URL(requestUrl);
+        return parsedUrl.protocol === 'https:';
+    } catch (err) {
+        throw new Error(`Invalid URL: ${requestUrl}. ${err.message}`);
+    }
 }
 
 enum EnvironmentVariables {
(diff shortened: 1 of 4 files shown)


lib/HttpClient.ts
```diff
@@ -84,13 +84,17 @@ export class HttpClientResponse implements ifm.IHttpClientResponse {
 
 export interface RequestInfo {
     options: http.RequestOptions;
-    parsedUrl: url.Url;
+    parsedUrl: URL;
```
tarunramsinghani: Do we understand the behaviour difference between oldurl.parse and new URL usage ? ANy specific kind of URL's that worked before will work now as well?

rajmishra1997: These changes are not required for UseNode test error, shall we keep the PR to remove deprecated use of url.parse ?

tarunramsinghani (approved):

lib/HttpClient.ts
rajmishra1997: Updated PR with all copilot suggestions

lib/Util.ts
```diff
@@ -25,21 +25,72 @@ export function getUrl(resource: string, baseUrl?: string, queryParams?: IReques
...
+                    mergedUrl.protocol = base.protocol;
+                }
+                if (!mergedUrl.host && base.host) {
+                    mergedUrl.host = base.host;
+                }
+                resultantUrl = mergedUrl;
+            } else {
+                // Resource is a relative path - merge with base using path.resolve
+                const protocol = base.protocol;
+                const encodedUsername = base.username ? encodeURIComponent(base.username) : '';
```
sanjuyadav24: username and password here can contain special characters?

rajmishra1997: Yes, urlstores username/password in decoded form (e.g., %40 becomes @). When we reconstruct the URL string manually, we must use 
 encodeURIComponent() to re-encode special character

lib/HttpClient.ts
```diff
@@ -593,7 +626,7 @@ export class HttpClient implements ifm.IHttpClient {
         return { proxyUrl: proxyUrl, proxyAuth: proxyAuth };
     }
 
-    private _isMatchInBypassProxyList(parsedUrl: url.Url): Boolean {
+    private _isMatchInBypassProxyList(parsedUrl: URL): Boolean {
```
raujaiswal, line 629: can we please check if we can use boolean here that is primitive type instead of object Type Boolean?

lib/HttpClient.ts
raujaiswal, line 4: can we please check if we need this url import in this file and other file since we removed url.parse?

PR #396, opened 2026-04-23
Updated package dependencies to fix vulnerabilities

sanjays-ms:
[AB#2362473](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2362473)

Updated package dependencies to fix vulnerabilities

Serialize-javascript package is a transitive dependency of the mocha package
Mocha version 11 doesn't have a fix for this and this fix is implemented in mocha v12.
However v12 is in beta so temporarily we have opted for an override.
Eventually once mocha v12 is released it will be replaced in the repository and the override removed.

Testing.
Only testing done through CI checks and running test cases locally.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typed-rest-client",
-  "version": "2.3.0",
+  "version": "2.4.0",
```
tarunramsinghani: Since this is only dependency change and no new changes, lets keep it 2.3.1 instead of 2.4.0

sanjays-ms: updated version to 2.3.1 as suggested.

tarunramsinghani (approved):

PR #397, opened 2026-05-11
Fix: Cast proxy port to Number to prevent CONNECT hang under Node 24

v-schhabra:
**Description:**
Cast proxy port to number to prevent tunnel hang on Node 24.
URL.parse().port returns a string (e.g. "8443"), but the tunnel library expects a number. In Node 24, passing a string port causes the HTTP. CONNECT handshake to silently hang until the task timeout (60 minutes), breaking all proxy-tunneled requests on self-hosted agents behind authenticated proxies.
https://github.com/microsoft/azure-pipelines-extensions/pull/1375

**Fix:**
 lib/HttpClient.ts (line 515):
 port: Number(proxy.proxyUrl.port)
 Casts the string port to a number before passing it to the tunnel library.

### Testing
- Added unit test verifying proxy port is numeric

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -512,7 +512,7 @@ export class HttpClient implements ifm.IHttpClient {
                 proxy: {
                     proxyAuth: proxy.proxyAuth,
                     host: proxy.proxyUrl.hostname,
-                    port: proxy.proxyUrl.port
+                    port: Number(proxy.proxyUrl.port)
                 },
             };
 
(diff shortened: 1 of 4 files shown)


rm-dmiri (approved):

lib/HttpClient.ts
```diff
@@ -512,7 +512,7 @@ export class HttpClient implements ifm.IHttpClient {
                 proxy: {
                     proxyAuth: proxy.proxyAuth,
                     host: proxy.proxyUrl.hostname,
-                    port: proxy.proxyUrl.port
+                    port: Number(proxy.proxyUrl.port)
```
tarunramsinghani: url.parse is deprecated in node24, we should update the use of url.parse

tarunramsinghani: lets merge https://github.com/microsoft/typed-rest-client/pull/389 before this PR

v-schhabra: > url.parse is deprecated in node24, we should update the use of url.parse

I have created new PR by replacing url.parse with URL

PR #399, opened 2026-05-12
Replace deprecated url.parse() with WHATWG URL constructor

v-schhabra:
**Description**

Migrates all usages of the legacy url.parse() API to the modern URL constructor (WHATWG URL Standard). url.parse() has been deprecated since Node 11 and emits deprecation warnings in Node 22+. It will be removed in a future Node.js version.

## Changes

- **lib/HttpClient.ts**: Replace all url.parse() calls with new URL(). Ensure proxy tunnel port is always numeric with a safe default when port is absent (prevents NaN/empty string for default ports).
- **lib/Interfaces.ts**: Change IRequestInfo.parsedUrl type from url.Url to the global URL type. Remove unused url import.
- **lib/Util.ts**: Rewrite getUrl() to use new URL(resource, base) with trailing-slash normalization so relative paths append correctly (matching prior path.resolve behavior). Remove url and path imports.
- **test/units/utiltests.ts**: Update test helpers to use new URL().

## Why

- url.parse() is deprecated and will be removed in future Node versions
- The URL constructor is the official replacement, stable since Node 10
- This ensures forward compatibility with Node 24+ and beyond

## Testing

- All 138 existing tests pass (82 unit + 56 integration)

## Breaking Change Assessment

IRequestInfo.parsedUrl type changed from url.Url to URL. Any consumers accessing properties like .auth, .query (as string), or .path on parsedUrl will need to update to the WHATWG URL equivalents (.username/.password, .searchParams, .pathname + .search). This is a minor semver bump.

### DIFF
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -88,7 +88,12 @@ export interface RequestInfo {
 }
 
 export function isHttps(requestUrl: string) {
-    let parsedUrl: URL = new URL(requestUrl);
+    let parsedUrl: URL;
+    try {
+        parsedUrl = new URL(requestUrl);
+    } catch (e) {
+        throw new Error(`Invalid URL "${requestUrl}": ${e.message}`);
+    }
     return parsedUrl.protocol === 'https:';
 }
 
@@ -243,7 +248,12 @@ export class HttpClient implements ifm.IHttpClient {
             throw new Error("Client has already been disposed.");
         }
 
-        let parsedUrl = new URL(requestUrl);
+        let parsedUrl: URL;
+        try {
+            parsedUrl = new URL(requestUrl);
+        } catch (e) {
+            throw new Error(`Invalid URL "${requestUrl}": ${e.message}`);
+        }
         let info: RequestInfo = this._prepareRequest(verb, parsedUrl, headers);
 
         // Only perform retries on reads since writes may not be idempotent.
@@ -294,7 +304,12 @@ export class HttpClient implements ifm.IHttpClient {
                     // if there's no location to redirect to, we won't
                     break;
                 }
-                let parsedRedirectUrl = new URL(redirectUrl, parsedUrl.href);
+                let parsedRedirectUrl: URL;
+                try {
+                    parsedRedirectUrl = new URL(redirectUrl, parsedUrl.href);
+                } catch (e) {
+                    throw new Error(`Invalid redirect URL "${redirectUrl}": ${e.message}`);
+                }
                 if (parsedUrl.protocol == 'https:' && parsedUrl.protocol != parsedRedirectUrl.protocol && !this._allowRedirectDowngrade) {
                     throw new Error("Redirect from HTTPS to HTTP protocol. This downgrade is not allowed for security reasons. If you want to allow this behavior, set the allowRedirectDowngrade option to true.");
                 }
@@ -581,7 +596,11 @@ export class HttpClient implements ifm.IHttpClient {
         let proxyAuth: string;
         if (proxyConfig) {
             if (proxyConfig.proxyUrl.length > 0) {
-                proxyUrl = new URL(proxyConfig.proxyUrl);
+                try {
+                    proxyUrl = new URL(proxyConfig.proxyUrl);
+                } catch (e) {
+                    throw new Error(`Invalid proxy URL "${proxyConfig.proxyUrl}": ${e.message}`);
+                }
             }
 
             if (proxyConfig.proxyUsername || proxyConfig.proxyPassword) {
--- a/lib/HttpClient.ts
+++ b/lib/HttpClient.ts
@@ -83,7 +83,8 @@ export class HttpClientResponse implements ifm.IHttpClientResponse {
 
 export interface RequestInfo {
     options: http.RequestOptions;
-    parsedUrl: URL;
+    parsedUrl: ifm.ILegacyParsedUrl;
+    requestUrl?: URL;
     httpModule: any;
 }
 
@@ -254,7 +255,7 @@ export class HttpClient implements ifm.IHttpClient {
         } catch (e) {
             throw new Error(`Invalid URL "${requestUrl}": ${e.message}`);
         }
-        let info: RequestInfo = this._prepareRequest(verb, parsedUrl, headers);
+        let info: RequestInfo = this._prepareRequest(verb, parsedUrl, headers, requestUrl);
 
         // Only perform retries on reads since writes may not be idempotent.
         let maxTries: number = (this._allowRetries && RetryableHttpVerbs.indexOf(verb) != -1) ? this._maxRetries + 1 : 1;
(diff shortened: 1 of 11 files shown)


tarunramsinghani (approved):

lib/Interfaces.ts
```diff
@@ -37,7 +36,7 @@ export interface IHttpClientResponse {
 
 export interface IRequestInfo {
     options: http.RequestOptions;
-    parsedUrl: url.Url;
+    parsedUrl: URL;
```
nemanjarogic, line 39: Since this is a public interface Any downstream consumer accessing properties like .`auth`, `.query` (as string), `.path`, or `.href` on parsedUrl will break. The WHATWG URL object has a different shape:

- url.Url.auth → no direct equivalent (use .username/.password)
- url.Url.path → gone (use .pathname + .search)
- url.Url.query (string) → .searchParams (object)
- url.Url.href behaved differently for partial URLs
- 
This is a breaking change for the public API and should warrant a major version upgrade, not the 2.3.1 → 2.3.2 patch bump in the PR.

nemanjarogic: Also consider potential effects of such breaking change

lib/HttpClient.ts
```diff
@@ -84,12 +83,12 @@ export class HttpClientResponse implements ifm.IHttpClientResponse {
...
 export interface RequestInfo {
     options: http.RequestOptions;
-    parsedUrl: url.Url;
+    parsedUrl: URL;
     httpModule: any;
 }
 
 export function isHttps(requestUrl: string) {
-    let parsedUrl: url.Url = url.parse(requestUrl);
+    let parsedUrl: URL = new URL(requestUrl);
```
nemanjarogic: `new URL()` throws on invalid/relative URLs, while `url.parse` never threw.  This could crash callers who previously relied on the forgiving behavior.

Worth checking: https://kilianvalkhof.com/2024/javascript/the-problem-with-new-url-and-how-url-parse-fixes-that/

The same comment goes for other places where we have `new URL()`

lib/HttpClient.ts
```diff
@@ -244,7 +243,7 @@ export class HttpClient implements ifm.IHttpClient {
             throw new Error("Client has already been disposed.");
         }
 
-        let parsedUrl = url.parse(requestUrl);
+        let parsedUrl = new URL(requestUrl);
```
nemanjarogic: Consider adding tests for malformed URL inputs since `new URL()` throws

lib/Interfaces.ts
v-schhabra: > Since this is a public interface Any downstream consumer accessing properties like .`auth`, `.query` (as string), `.path`, or `.href` on parsedUrl will break. The WHATWG URL object has a different shape:
> 
> * url.Url.auth → no direct equivalent (use .username/.password)
> * url.Url.path → gone (use .pathname + .search)
> * url.Url.query (string) → .searchParams (object)
> * url.Url.href behaved differently for partial URLs
> 

I have addressed this by introducing ILegacyParsedUrl — a backward-compatible interface that preserves all the legacy properties (.auth, .path, .query, .host, .port, .search, .hash) with the same semantics as url.Url. The IRequestInfo.parsedUrl field now uses this type, so downstream consumers won't break.

v-schhabra: > Also consider potential effects of such breaking change

I built a _toLegacyParsedUrl() adapter that maps WHATWG URL properties back to the legacy shape, handling edge cases like:
  - Explicit default ports (:80/:443) that new URL() strips
  - Encoded characters in auth (%40, %3A, %2F) properly decoded
  - Empty ? and # markers preserved in .href
  - IPv6 hosts with port
 And we validated this with a comprehensive integration test that asserts all legacy fields match url.parse() behavior for a complex URL with auth, port, query, and fragment.

lib/HttpClient.ts
v-schhabra: I have implemented the URL.parse() pattern from this article. Since URL.parse() requires Node 22.1+ 
 we use a fallback:
- If URL.parse is available → use it (returns null, no throw)
- Otherwise → try/catch around new URL() with a descriptive error message

lib/HttpClient.ts
v-schhabra: Done as suggested.

lib/Interfaces.ts
nemanjarogic: Do you think this is worth effort? This reconstructs all legacy url.Url properties from a WHATWG URL. It's technically sound and solves the problem I mentioned. However, it adds significant maintenance burden, so I am worried about this part. 

Maybe we can ask a few other folks for opinion, but wouldn't it be long-term more reasonable to go update major version and go with breaking change? In the end, legacy things are deprecated and with new versions we should migrate to the newly supported infrastructure, not necessarily migrate and then maintain old things.

nemanjarogic: One reason for that is that this code is also surface for bugs.

For example, the adapter uses `getRawHost(rawRequestUrl) `to recover explicit default ports (:80, :443) that WHATWG URL strips. For the initial request, `rawRequestUrl` is the user's original string, so this works. For redirects, rawRequestUrl is `parsedRedirectUrl.href` - already WHATWG-normalized with default ports stripped. So _toLegacyParsedUrl can't recover them for redirect URLs

lib/HttpClient.ts
nemanjarogic: I don't see any usage of URL.parse(). Can you show me where it's used?

v-schhabra: Sorry I was planning to add that but due to node version constraint I didnt do that.

lib/HttpClient.ts
```diff
@@ -449,6 +465,76 @@ export class HttpClient implements ifm.IHttpClient {
         return info;
     }
 
+    private _toLegacyParsedUrl(requestUrl: URL, rawRequestUrl?: string): ifm.ILegacyParsedUrl {
```
tarunramsinghani: this seems like we are carrying the debt from old url instead of moving forward. the typed rest client is library and if anybody uses the new version they should adjust to new behaviour. We should go to newer ebhaviour instead of creating the new interface

lib/Interfaces.ts
tarunramsinghani: I agree that we should not try to build a legacy layer, I think we can ship with new behaviour with major version upgrade.

v-schhabra: Done as suggested.

lib/opensource/Node-SMB/lib/ntlm.js
```diff
@@ -217,7 +217,13 @@ exports.challengeHeader = function (hostname, domain) {
 
 exports.responseHeader = function (res, url, domain, username, password) {
   var serverNonce = Buffer.from((res.headers['www-authenticate'].match(/^NTLM\s+(.+?)(,|\s+|$)/) || [])[1], 'base64');
-  var hostname = require('url').parse(url).hostname;
+  var parsedUrl;
+  try {
```
nemanjarogic: Let's also abandon try/catch around `new URL` since we agreed we are fine with breaking changes

nemanjarogic: Please apply this across all files and places which follow the same pattern

lib/HttpClient.ts
v-schhabra: Sure

nemanjarogic (approved):

tarunramsinghani (approved):

PR #258, opened 2021-03-29
Updating Underscore to resolve SNYK-JS-UNDERSCORE-1080984

jessehouwing:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/typed-rest-client/258)

PR #414, opened 2026-09-15
Drop the underscore dependency from the NTLM handler

### LINKED ISSUE #413: Drop underscore after two CVE updates for functions not called here
## Symptom

This repository has made two underscore security updates for vulnerabilities in functions typed-rest-client does not call:

- **CVE-2021-23358** (critical, arbitrary code execution through the template function): #258 updated underscore to 1.12.1, and #261 linked that fix. #284 later reported the same advisory through `azure-pipelines-tasks-azure-arm-rest-v2`; microsoft/azure-pipelines-tasks#15087 updated that downstream package's typed-rest-client dependency.
- **CVE-2026-27601** (high, unbounded recursion in `_.flatten` and `_.isEqual`): #392 updated the package dependency to 1.13.8 on March 10.

At master @ 0ad6895 the only underscore calls are `_.extend` and `_.omit`; there is no static reference to `_.template`, `_.flatten` or `_.isEqual` anywhere in the tree. Both advisories therefore prompted dependency updates for functions typed-rest-client does not call. Package version audits such as `npm audit` surface those advisories through typed-rest-client in downstream dependency trees, while function use stays a separate mitigating factor check. Removing underscore removes that dependency level exposure.

I noticed this while investigating a downstream [socket.dev](https://socket.dev) [`Obfuscated code`](https://socket.dev/alerts/obfuscatedFile) heuristic on `underscore@1.13.8`; that alert is not the reason for this request.

## Code path (master @ 0ad6895)

`underscore` is imported by one file, for six calls on four lines:

1. [`lib/handlers/ntlm.ts:16`](https://github.com/microsoft/typed-rest-client/blob/0ad6895a218041801cfe5b468454566e828757de/lib/handlers/ntlm.ts#L16) imports it.
2. [`ntlm.ts:77`](https://github.com/microsoft/typed-rest-client/blob/0ad6895a218041801cfe5b468454566e828757de/lib/handlers/ntlm.ts#L77) `_.extend` merges the credentials into the request options.
3. [`ntlm.ts:129`](https://github.com/microsoft/typed-rest-client/blob/0ad6895a218041801cfe5b468454566e828757de/lib/handlers/ntlm.ts#L129) `_.extend(type1options, _.omit(requestInfo.options, 'headers'))` builds the type 1 request.
4. [`ntlm.ts:183-184`](https://github.com/microsoft/typed-rest-client/blob/0ad6895a218041801cfe5b468454566e828757de/lib/handlers/ntlm.ts#L183-L184) does the same for the type 3 request, plus an `_.extend` of the caller's headers.

Both functions walk every enumerable string key, own or inherited, and skip symbols. That is not `Object.assign`, which copies own keys and symbols and skips inherited ones: a custom `prepareRequest` handler runs over the same options object before NTLM does and can put options on its prototype, and with `Object.assign` an inherited `signal` stopped cancelling the handshake in a local test. For the object valued options and headers at these sites, a `for...in` helper preserves underscore's enumerable string key domain.

## Suggested fix

A short `assignKeys(target, source, skip?)` helper replaces all six calls, `underscore` leaves `dependencies`, and its entry leaves `ThirdPartyNotice.txt`. The change is on a branch with a unit test that drives the three leg handshake through nock with an earlier handler setting an inherited option; it checks that option on all three requests, a caller header on the initial and type 3 requests, and that the credentials land on the same options object the handlers were given. Replacing `for...in` with own key iteration makes it fail, as does merging into a copy.

#157 considered the same removal in 2019 and kept underscore because changing NTLM was high risk and time consuming to test relative to a 112 KB size saving. The two later security updates add a maintenance consideration that was not part of that tradeoff, and the proposed test makes the NTLM behaviour measurable. I can open the pull request if that would be useful.

cplieger:
`underscore` is imported by one file, `lib/handlers/ntlm.ts`, for six calls on four lines: four `_.extend` and two `_.omit(options, 'headers')`. This replaces them with a small `assignKeys` helper and drops the dependency. Motivation and the trace are in the linked issue: this repository has made two underscore security updates (#258 for CVE-2021-23358, #392 for CVE-2026-27601) for vulnerabilities in functions it does not call.

## Why `for...in` and not `Object.assign`

Both underscore functions walk every enumerable string key, own or inherited, and skip symbols. `Object.assign` copies own keys and symbols and skips inherited ones. The stock `HttpClient` path only ever produces own string keys, but a custom `prepareRequest` handler runs over the same options object before NTLM does and can put options on its prototype. In a local test with `Object.assign`, an inherited `signal` stopped cancelling the handshake, and an enumerable symbol getter on the headers was invoked where underscore ignored it. `for...in` preserves the key domain underscore had for the object valued options and headers that reach these sites.

## Test

The new `[NTLM]` case drives the three leg handshake through nock with an earlier handler that sets an inherited option and a caller header. It checks the inherited option on all three requests, the caller header on the initial and type 3 requests, and that NTLM's credentials land on the options object the handlers were given rather than on a copy.

I checked it against six single mutants, one at a time, and it fails under each: the helper keeping headers, the credentials merged into a fresh copy, the type 3 header merge removed, own key iteration, the skip ignored, and the helper returning its source.

## ThirdPartyNotice.txt

Underscore's index entry and license block are removed by hand. `generate-third-party-notice.js` does not run against a current npm layout (it reads `node_modules/.package-lock.json` as a package directory), so hand editing is how the file has been maintained. The `Based on Underscore` lines inside lodash's own notices are lodash's and stay.

Closes #413


cplieger:
@microsoft-github-policy-service agree

